PR #1190, opened 2014-04-24
Allow creating basic tier VMs

mhyllander:
To allow creating Basic tier VM instances, I changed the parsing of the vm size argument passed to "vm create", to allow specifying basic_a0, basic_a1, basic_a2, basic_a3 and basic_a4.


mhyllander:
I believe this pull request is now unnecessary, the current implementation solves the problem by allowing any VM size.

PR #1473, opened 2014-12-08
Update dependencies

bear454:
The following updates allow azure-cli to better follow current versions of dependent projects while currently showing no impact on the test suite.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "request": "~2",
     "streamline": "~0.4",
     "through": "2.3.4",
-    "tunnel": "0.0.2",
+    "tunnel": "~0.0.2",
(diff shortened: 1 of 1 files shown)


amarzavery:
We will not take this pull request because some of the package versions break the current version of CLI.
For example streamline and request packages.

bear454:
None of these changes caused test failures at the time they were submitted.

Is it going to be the general policy of this project to ignore bugfixes, performance improvements, and new features in upstream node modules? If so, packaging this in any meaningful way is impossible.

amarzavery:
We are not against bugfixes, improvements and new features in upstream node modules.
Requirements for some of the packages are heavily relaxed which is problematic.

PR #1494, opened 2014-12-18
Fix login issues in AzureChinaCloud

markcowl:


yugangw-msft:
we will fix in anther pr, closing

PR #1525, opened 2015-01-13
Support storage stored access policy and update storage SDK

yaxia:
1. Updated storage SDK to 0.4.1.
2. Supported stored access policy for blob, table and queue services.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/util/storage.util._js
```diff
@@ -738,19 +748,265 @@ StorageUtil.isValidRetention = function(value) {
...
+  var policies = [];
+  StorageUtil.startProgress(policySettings.tips);
+  try {
+    policies = StorageUtil.performStorageOperation(policySettings.getAclOperation, _, policySettings.resourceName, policySettings.storageOptions).signedIdentifiers;
+
+    var newPolicy = generatePolicy(policySettings.policyName, policySettings.policyOptions.start, policySettings.policyOptions.expiry, policySettings.policyOptions.permissions);
+    modifyPolicySet(newPolicy, policies, policySettings.accessType, operation);
+
+    if (policySettings.accessType === StorageUtil.AccessType.Container) {
+      StorageUtil.performStorageOperation(policySettings.setAclOperation, _, policySettings.resourceName, policies, null, policySettings.storageOptions);
```
yugangw-msft: Just  questiuon that  you do mean to pass the "_" (streamline implict callback) in the middle, rather usually the last argument

lib/util/storage.util._js
```diff
@@ -738,19 +748,265 @@ StorageUtil.isValidRetention = function(value) {
...
+
+/**
+* Modify the sotrage policies set that would be sent
+*/
+function modifyPolicySet(newPolicy, existingPolicies, accessType, operation) {
+  var index = -1;
+  for (var i = 0; i < existingPolicies.length; i++) {
+    normalizePolicy(existingPolicies[i], accessType);
+    if (existingPolicies[i].Id === newPolicy.Id) {
+      index = i;
```
yugangw-msft: need to break here?

lib/util/storage.util._js
```diff
@@ -738,19 +748,265 @@ StorageUtil.isValidRetention = function(value) {
...
+    }
+  }
+
+  if (operation == StorageUtil.PolicyOperation.Create) {
```
yugangw-msft: nit, use ===.?

lib/util/storage.util._js
yaxia: This is to follow the parameter pattern of StorageUtil.performStorageOperation(storageOperation, _). "_" is required as second parameter in the function declaration, So the parameters for "storageOperation" are behind "_".

lib/util/storage.util._js
yaxia: The available index is set only when the "if" clause is true, but normalizePolicy() should be applied to all the items. So we don't break here.
And the maximum policy number is 5, so the redundant checking won't have much impact.

lib/util/storage.util._js
yaxia: Resolved. Thanks.

PR #1534, opened 2015-01-15
Disable blob cases for pending investigation for failures on node 0.8

yaxia:
As we met "Error: timeout of 500000ms exceeded" from time to time which was caused by stream issue in nodejs 0.8. So I disable the cases until we update storage nodejs SDK to 0.4.3.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

PR #1555, opened 2015-01-28
updated mocks with url changes and updated "delete" to "deleteMethod" as it is a reserved keyword in JS

amarzavery:
updated the mock recordings with url changes and updated method name from "delete" to "deleteMethod" because of the changes in hyak specs for reserved keywords in JS.


amarzavery:
merging the changes. the tests will pass in azure-sdk-for-node repo once the corresponding node sdk changes are merged.

Issue #1523, opened 2015-01-11
[proposal] Set the CN for the default docker certification to *.cloudapp.net


waitingkuo:
The proposal is to modify the Common Name of the default docker certification and solve #1493 

Currently, the CN we signed is *, it doesn't match the default DNS azure privided "mydomain.cloudapp.net"

I currently add a record in /etc/hosts to fix that.
If we modify the CN to *.cloudapp.net, then i'll match without modifying our /etc/hosts.

kmouss:
This is dup of this: https://github.com/Azure/azure-xplat-cli/issues/1493

Issue #1493, opened 2014-12-18
Failed to connect docker deamon with --tlsverify option


phstudy:
The default certificate generated by `azure vm docker create ...` command will use `*` as `CN`, which does not work with `--tlsverify` option.

The default certificate works with `--tls` option
~~~
17:52:03 study-mbpr: azure-deploy$ docker --tls -H tcp://phstudy.azure.tw:4243 ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
~~~

The default certificate does not work with `--tlsverify` option
~~~
17:52:11 study-mbpr: azure-deploy$ docker --tlsverify -H tcp://phstudy.azure.tw:4243 --tlscacert="/Users/study/.docker/ca.pem" --tlscert="/Users/study/.docker/cert.pem" --tlskey="/Users/study/.docker/key.pem" ps
2014/12/18 17:52:19 Get https://phstudy.azure.tw:4243/v1.15/containers/json: x509: certificate is valid for *, not phstudy.azure.tw
~~~

If I set environment variable `DOCKER_TLS_VERIFY=1`, I will also get the same error as above.
~~~
11:00:34 study-mbpr: ~$ export DOCKER_TLS_VERIFY=1
11:00:35 study-mbpr: ~$ docker --tls -H tcp://phstudy.azure.tw:4243 ps
2014/12/19 11:00:40 Get https://phstudy.azure.tw:4243/v1.15/containers/json: x509: certificate is valid for *, not phstudy.azure.tw
~~~

And there seems no way to change default certificate in created docker vm, it will be replaced with the original one after rebooting.

ahmetalpbalkan:
@phstudy your domain name does not have a SSL certificate. You need to purchase a SSL cert if you want to use --tlsverify. http://azure.microsoft.com/en-us/documentation/articles/cloud-services-configure-ssl-certificate/

Does this solve your problem?

phstudy:
@ahmetalpbalkan I know if I want to use --tlsverify,  my domain should have a SSL certificate. But another question is how to renew the certificate? If I put new certificate in docker vm, it will be replaced with the old certificate.

bureado:
[...]

PR #1560, opened 2015-02-01
Per-instance server certificate creation for azure vm docker create

bureado:
As referenced in #1493 and #1523, in order to use --tlsverify the server certificates created upon first initialization of .docker need to have a CN different than *, this change generates the ca and client keys and certs upon .docker initialization, and the server keys and certs every time an instance is created, with dnsName + 'cloudapp.net' as CN.

### DIFF
--- a/lib/commands/asm/vm/vmclient.js
+++ b/lib/commands/asm/vm/vmclient.js
@@ -3370,7 +3370,7 @@ function createVM(options, callback, logger, cli) {
 
       progress = cli.interaction.progress($('Deleting cloud service'));
 
-      computeManagementClient.hostedServices.delete(options.dnsPrefix, function(err) {
+      computeManagementClient.hostedServices.deleteMethod(options.dnsPrefix, function(err) {
         progress.end();
         if (err) {
           logger.warn(util.format($('Error deleting %s cloud service'), options.dnsPrefix));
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@bureado__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

hopetobelievein:
@bureado When your command creates server certificates it uses <code>dnsName + '.cloudapp.net'</code> pattern for CN:
https://github.com/bureado/azure-xplat-cli/blob/master/lib/commands/asm/vm/vmclient.js#L3920

For example in issue https://github.com/Azure/azure-xplat-cli/issues/1493 customer has <code>*.azure.tw</code> DNS. Will this fix work in this case?

bureado:
I don't know if azure.tw, azure.cn or other Azure clouds that xplat-cli can work with use cloudapp.net or a different suffix.

If this were the case then I'd suggest the default suffix is part of the subscription information or is tied to the VM creation location via a hash to avoid hitting Azure for that information. In fact I don't know if the default suffix is tied to locations or subscriptions.

Note this patch tries to fix two things: 1) a real hostname that can then be used for TLS verification and 2) separate certificates for multiple Docker VMs. The latter is irrespective of the suffices.

ahmetalpbalkan:
Is there a way to get the full DNS name and not hardcode `.cloudapp.net`? @bureado

hovsepm:
Please fix the build breaks. 
Please update this PR by EOW. We will close it otherwise.

ahmetalpbalkan:
@bureado  ping, do you want to get this in?

PR #1590, opened 2015-02-18
Integrate from Release 0.8.15 to master

yugangw-msft:


yugangw-msft:
0.8 and 0.12 failures are known issues which will be addressed in the upcoming in the sprint. Merging

PR #1596, opened 2015-02-20
[do not merge] upgrade streamline version so test could run on node 0.12.0

yugangw-msft:


amarzavery:
Figure out why 0.12 takes 16 mins vs 0.10 2.28 mins

PR #1620, opened 2015-03-09
Fix for massive verbose output during test run (repro'd only on node 0.12.0)

yugangw-msft:
We will need to fix it. It makes test run longer and the log is unreadable.
The root cause is on the kuduscript package, which one negative test hits a streamline bug from it.
The fix for the package owner is to migrate streamline dependency to the latest version which contains the fix See https://github.com/Sage/streamlinejs/issues/242


yugangw-msft:
reviewed by amar

PR #1624, opened 2015-03-10
fix installer issues in mac and windows

yugangw-msft:
This is caused by our streamline package upgrading needed for node 0.12.0


yugangw-msft:
failed on node 0.8 for mobile test flakiness, which I fixed in the dev branch today

PR #1627, opened 2015-03-11
Improve the windows installer script  

yugangw-msft:
1. Check errorlevel correctly, what we have now is not correct. 
2. Checkout exit code of each step consistently, including node.exe and robocopy.exe
3. Stop the script if there are any errors, and do not continue
4. Check folder exists before delete, and doing this will get rid of all those output of "file doesn't exist"
5. set codepage, so "npm install" works consistently

### DIFF
--- a/tools/windows/scripts/prepareRepoClone.cmd
+++ b/tools/windows/scripts/prepareRepoClone.cmd
@@ -35,7 +35,10 @@ echo Temporary clone of the repo already exists. Removing it...
 pushd %TEMP_REPO%\..\
 if exist %TEMP_REPO_FOLDER% rmdir /s /q %TEMP_REPO_FOLDER%
 ::rmdir always returns 0, so check folder's existence 
-if exist %TEMP_REPO_FOLDER% goto ERROR
+if exist %TEMP_REPO_FOLDER% (
+    echo Failed to delete %TEMP_REPO_FOLDER%.
+    goto ERROR
+)
 popd
 
 :CLONE_REPO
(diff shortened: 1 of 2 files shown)


tools/windows/scripts/prepareRepoClone.cmd
```diff
@@ -6,65 +6,62 @@
...
-REM echo Creating output directory 'out'...
-REM mkdir .\out
-IF NOT ERRORLEVEL 0 GOTO ERROR
+set OUTPUT_FOLDER=.\out
+if exist %OUTPUT_FOLDER% rmdir /s /q %OUTPUT_FOLDER%
+mkdir %OUTPUT_FOLDER%
 
+set TEMP_REPO_FOLDER=azure-cli
+set TEMP_REPO=%temp%\%TEMP_REPO_FOLDER%
+if not exist %TEMP_REPO% goto CLONE_REPO
```
amarzavery: Wow!! goto statements are still in use :)

amarzavery:
LGTM

tools/windows/scripts/prepareRepoClone.cmd
```diff
@@ -6,65 +6,62 @@
...
 
-
 :CLONE_REPO
 mkdir %TEMP_REPO%
 echo Cloning the repo elsewhere on disk...
 pushd ..\..\
 robocopy . %TEMP_REPO% /MIR /XD .git tools features scripts test node_modules /NFL /NDL /NJH /NJS
-IF NOT ERRORLEVEL 0 GOTO ERROR
-echo.
+if %errorlevel% geq 8 goto ERROR
```
amarzavery: one question out of curiosity ... I want to understand why do we go to error if the level is greater than or equal to 8 in this case and 
go to Error if it is not equal to 0 in other cases in the script?

yugangw-msft: Robocopy communicates an error using exit value (>=8), while most other commands use any non-zero value including "node.exe". 
Let me add a comment here, since yesterday i was badly confused

PR #1658, opened 2015-03-28
removed sqlserver rp registration

amarzavery:


yugangw-msft:
let me close this one, as i will submit a new one with deeper cleaning :-)

PR #1659, opened 2015-03-30
Do not register providers unless when they are indeed needed

yugangw-msft:
So I removed the old logic and now only register when a command needs it. We will suggest this to any partner teams when they on-board new commands.
This change did some more cleaning up. The old one is pretty bogus, and behaves random, because it does the registration even the current commands do not need it. So, 
1. For ARM providers such as "cleardb", I simply removed them from the registration list, because they will be covered when the deployment happens. 
2. For RDFE providers, sqlserver is removed from the list, because it is always registered. WebSite will be registered only when the client is created by related commands.

I also consolidated the way of how an azure client gets created. The current logic is unnecessary complicated, and inconsistent, for example some clients are created by subscription.js but more are done by "utils.js", worse, the "subscription.js" also delegates some to the "utils.js". For the sake of readability, my change addresses them:
1. Flatten the logic of creating the client. The "client creating" will only be handled at "utils.js", "subscription.js" no longer expose the api for creating
2. Delete parameters which were passed around, but actually never used, such as "log"
3. Fix the function name, particularly those ones with underscore in the naming but are actually public and used by external files.


lib/util/utils.js
```diff
@@ -59,27 +56,42 @@ exports.createClient = function (factoryOrName, credentials, endpoint) {
...
+                              subscription._createCredentials(), 
+                              subscription.resourceManagerEndpointUrl);
+}
 
-exports._createWebsiteClient = function (subscription, logger) {
-  return exports._createService('createWebSiteManagementClient', subscription, logger);
+//website provider might not be registered yet, so make sure register it
+exports.createWebsiteClient = function (subscription, callback) {
+  var client;
+  if (!subscription.registeredProviders.some(function (item) {
```
amarzavery: move detection logic inside subscription.js and rename "registerAsmProvider" to "ensureAsmProviderRegistration"

lib/util/utils.js
```diff
@@ -59,27 +56,42 @@ exports.createClient = function (factoryOrName, credentials, endpoint) {
...
-};
+function _createArmClient(factory, subscription) {
+  return exports.createClient(factory,
+                              subscription._createCredentials(), 
+                              subscription.resourceManagerEndpointUrl);
+}
 
-exports._createWebsiteClient = function (subscription, logger) {
-  return exports._createService('createWebSiteManagementClient', subscription, logger);
+//website provider might not be registered yet, so make sure register it
```
amarzavery: sure "to" register

PR #1665, opened 2015-03-31
Optimization of RP registration

amarzavery:
- Made ARM mode RP registration commands public
- Brought ASM mode RP registration commands in working condition
- Re-factored provider registration logic, to make things simpler


yugangw-msft:
lgtm

PR #1675, opened 2015-04-06
Do not cache registered namespace in local .azure\azureprofile.json

yugangw-msft:
Rather detect the error of 409 and give out an instruction which tell users to execute provider register command. Caching is generally disliked, because of not simple coding logic, command parallel running scenario support, etc.
Also modify"provider [register/unregister]" commands to polling the provider state to ensure the command went through


yugangw-msft:
I tried and decided not to write an extra test for verifying the error message. The reason is that "unregistering a provider" could take a very long time, and the command might exit with a warning the provider still staying at "unregistering", when the test continues to create a resource, it will land up with a different  error. I can hack it up to get the test going, but it will sure cause confusion when someone else tries to run the test live sometime later. Unregistering a provider is not a common scenario anyway.

lib/util/profile/subscription.js
```diff
@@ -263,15 +217,33 @@ _.extend(Subscription.prototype, {
...
-        });
+      self._waitForRegistrationComplete(client, false, numRetries, namespace, function (err) {
+        if (err) {
+          log.warn(util.format($('Unregistration is still ongoing. You can execute \'Azure Provider Show\' to monitor the status')));
+        }
+        callback();
+      });
+    });
+  },
+
```
amarzavery: Please document the function.
```
/**
 *
 *
*/
```

lib/util/profile/subscription.js
```diff
@@ -263,15 +217,33 @@ _.extend(Subscription.prototype, {
...
       }
-      //remove the namespace from the list of registered resource namespaces
-      if (self.registeredResourceNamespaces) {
-        self.registeredResourceNamespaces = 
-          self.registeredResourceNamespaces.filter(function (item) {
-            return (!utils.ignoreCaseEquals(item, namespace));
-        });
+      self._waitForRegistrationComplete(client, false, numRetries, namespace, function (err) {
+        if (err) {
+          log.warn(util.format($('Unregistration is still ongoing. You can execute \'Azure Provider Show\' to monitor the status')));
```
amarzavery: azure --> 'a' should be lower case

lib/util/utils.js
```diff
@@ -211,6 +212,19 @@ function createFollowRedirectFilter() {
 
 exports.createFollowRedirectFilter = createFollowRedirectFilter;
 
+polishErrorCausedByArmProviderNotRegistered = function () {
+  return function handle(resource, next, callback) {
+    var nextStream = next(resource, function (err, response, body) {
+      if (err && response && response.statusCode === 409) {
+        err.message = util.format('%s You can register it by executing the command of \'azure provider register\'.', 
+                                  err.message);
```
amarzavery: fix indentation

yugangw-msft: sorry, i saw it wrong. The indent is correct as it is the last function  parameter

lib/util/profile/subscription.js
yugangw-msft: fixed

lib/util/profile/subscription.js
yugangw-msft: added

PR #1677, opened 2015-04-07
update tests to works with VS mocha runner

yugangw-msft:


yugangw-msft:
Let us see whether we can make the test suite inherit from mocha base suite. If it turns out too costly, merge this one.

yugangw-msft:
reviewed and merge

PR #1690, opened 2015-04-12
Enhance helpJSON with package information

devigned:
Add some top level package information to the helpJSON representation.

### DIFF
--- a/lib/util/extendedcommand.js
+++ b/lib/util/extendedcommand.js
@@ -23,8 +23,8 @@ var commander = require('commander');
 var fs = require('fs');
 var path = require('path');
 var util = require('util');
-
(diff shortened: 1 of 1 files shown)


lib/util/extendedcommand.js
```diff
@@ -140,7 +140,22 @@ _.extend(ExtendedCommand.prototype, {
   },
 
   helpJSON: function() {
-    var result = {};
+    var self = this;
+    var result = _.tap({}, function(res){
+      if(_.isNull(self.parent) || _.isUndefined(self.parent)){
```
yugangw-msft: IMO, can we just check "!(self.parent)"

PR #1695, opened 2015-04-14
Wait for deployment to complete

vivsriaus:
The group deployment create command originally returned as soon as the create command was issued (where users would see a provisioningState of Accepted as soon as they hit enter). This is not the ideal, since users would not know if the deployment failed or not. We're trying to mirror the experience in powershell, where we wait for the deployment to succeed/fail and then call the displayDeployment(), instead of calling displayDeployment() as soon as the /create command as issued.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

yugangw-msft:
change looks good, but could you take a look to get CI passing before i can merge

amarzavery:
@vivsriaus - U might have to re-record other failing tests. Please look at the CI logs.

Tip: run ```node scripts\unit.js testlistarm.txt``` in plyaback mode on your machine to make sure all the ARM tests are passing.

vivsriaus:
@amarzavery @yugangw-msft Please not, including with this PR is a change to account command to display tenantId as well. The changes were minimal (2 lines), but please review them as well. Thanks

amarzavery:
Please add a deployment test without the return switch. You can also modify a current test to not use the -r switch. This is to ensure that all the functionality is well tested.

lib/commands/arm/group/group.deployment._js
```diff
@@ -42,6 +42,7 @@ exports.init = function (cli) {
     .option('-s --storage-account <storage-account>', $('the storage account where we will upload the template file'))
     .option('-p --parameters <parameters>', $('a JSON-formatted string containing parameters'))
     .option('-e --parameters-file <parametersFile>', $('a file containing parameters'))
+    .option('-r --return', $('returns as soon as deployment is created'))
```
yugangw-msft: i would say using --nowait is better. But i will merge it now since it contains the test fix which blocks several other PR, we can tweak it later.

PR #1696, opened 2015-04-15
Get 'token entry' encoding works when there is no refresh token (log in using a service principal)

yugangw-msft:
it is well known that under non interactive flow, adal will not give us the refresh token when xplat logs in as a service principal. So we need to handle this case


yugangw-msft:
the CI failure is not caused by my change.

PR #1698, opened 2015-04-16
Prompt for template params when parameters(obj) and parameters(file) are...

vivsriaus:
... not specified. Also change -r switch in group create deployment to --nowait

### DIFF
--- a/lib/commands/arm/group/groupUtils._js
+++ b/lib/commands/arm/group/groupUtils._js
@@ -365,7 +365,7 @@ function getTemplateParams(cli, subscription, options, _)
            options.galleryTemplate,
            _);
 
-  var paramKeys = new Object();
+  var paramKeys = {};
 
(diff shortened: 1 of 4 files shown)


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/commands/arm/group/groupUtils._js
```diff
@@ -265,6 +265,18 @@ function createDeploymentParameters(cli, subscription, resourceGroup, options, _
     deploymentParameters = JSON.parse(utils.stripBOM(jsonFile));
   } else if (options.parameters) {
     deploymentParameters = JSON.parse(options.parameters);
+  } else {
```
amarzavery: This option will come into action when the user does not provide anything (-p or -e) for parameters. Is my understanding correct?

vivsriaus: @amarzavery That is indeed correct

PR #1699, opened 2015-04-16
fix the hang when login using invalid service principal key

yugangw-msft:


lib/util/authentication/adalAuth.js
```diff
@@ -168,26 +168,47 @@ function loadServicePrincipalKey(appId, tenantId, callback) {
...
+              //will fail
+              removeInvalidServicePrincipalKey(self.appId, self.authConfig.tenantId, function(errorOnRemove){
+                if (errorOnRemove) {
+                  //this error would be rare, but warn still, just in case.
+                  logging.warn($('Failed to clean up invalid credentils for service principal'));
+                }
+                return callback(util.format($('Unable to acquire token due to error of: %s'), err.message));
+              });
+          } else {
+            return callback(null, 'Bearer', result.accessToken);
```
amarzavery: result.tokenType

yugangw-msft:
VM test flakiness on node 0.8, nothing to do with this PR. Have communicated with the owner, will track and fix

Issue #1688, opened 2015-04-10
group deployment show errors with "Invalid $filter 'provisioningState eq ''' specified in the query string"


paulmey:
This error comes directly from the API's:

```console
silly:   requestOptions
silly:   {
silly:       rawResponse: false,
silly:       queryString: {},
silly:       method: 'GET',
silly:       headers: { user-agent: 'WindowsAzureXplatCLI/0.8.17', Content-Type: 'application/json; charset=utf-8' },
silly:       url: "https://management.azure.com/subscriptions/c33f5e65-247c-4819-8bcc-0cb2c3b354df/resourcegroups/hack/deployments/?$filter=provisioningState%20eq ''&api-version=2014-04-01-preview"
silly:   }
silly:   returnObject
silly:   {
silly:       statusCode: 400,
silly:       header: {
silly:           x-ms-routing-request-id: 'WESTUS:20150410T212413Z:88362d40-ba3e-4229-90a8-1ed94851db7d',
silly:           date: 'Fri, 10 Apr 2015 21:24:13 GMT',
silly:           x-ms-correlation-request-id: '88362d40-ba3e-4229-90a8-1ed94851db7d',
silly:           strict-transport-security: 'max-age=31536000; includeSubDomains',
silly:           expires: '-1',
silly:           x-ms-request-id: '88362d40-ba3e-4229-90a8-1ed94851db7d',
silly:           pragma: 'no-cache',
silly:           x-ms-failure-cause: 'gateway',
silly:           x-ms-ratelimit-remaining-subscription-reads: '31998',
silly:           content-length: '136',
silly:           content-type: 'application/json; charset=utf-8',
silly:           cache-control: 'no-cache'
silly:       },
silly:       body: "{"error":{"code":"InvalidProvisioningStateFilter","message":"Invalid $filter 'provisioningState eq ''' specified in the query string."}}"
silly:   }
error:   Invalid $filter 'provisioningState eq ''' specified in the query string.
```

amarzavery:
closing this issue as the corresponding PR has been merged.

Issue #1692, opened 2015-04-13
azure role assignment list fails


yangl900:
Running 

'''
azure role assignment list
'''

returns:
error:   The filter 'principalId eq ''' is not supported. Supported filters are either 'atScope()' or 'principalId eq '{value}''.

the default role assignment listing should not pass in principalId filter. this seems a regression.

yugangw-msft:
Vivek, could you take a look? Come by if you need more context.

amarzavery:
I have almost completed this issue. Will send a PR shortly. Please review it.

amarzavery:
closing this issue as the corresponding PR has been merged.

PR #1703, opened 2015-04-17
do not provide empty string for filtering while showing or listing (resource/deployment)

amarzavery:
fixed issue #1688 and #1692


amarzavery:
#1688 and #1692

vivsriaus:
@amarzavery LGTM as long as the build passes :)

PR #1739, opened 2015-05-06
Perf improvement of xplat general command loading

yugangw-msft:
Using my lenovo carbon x1 laptop, this change improves the loading time from 9.1 s to 4.2 s, mainly by removing the dependency on azure roll-up package 
I also removed the code of plugin caching. The reason is the reloading has been disabled for an unresolved error since last September, so this feature has been useless. Also it charges 600 ms by writing on the file every time you run a command. So it is obvious that this code has to go. But, we can still pick it up if we can figure out the failure point, and also the plugin caching still provides good code snippets we can port.


lib/util/utils.js
```diff
@@ -113,67 +109,64 @@ exports.createWebSiteExtensionsClient = function (siteName, hostNameSuffix, user
...
+  var factoryMethod = require('./../commands/arm/temporary/storage').createStorageManagementClient;
+  return _createArmClient(factoryMethod, subscription);
 };
 
 exports.createComputeResourceProviderClient = function (subscription) {
-  var computeClient = require('./../commands/arm/armsdk/compute');
-  var functionName = 'createComputeManagementClient';
-  azure[functionName] = computeClient.createComputeManagementClient;
-  return _createArmClient(functionName, subscription);
+  var factoryMethod = require('./../commands/arm/armsdk/compute').createComputeManagementClient;
```
amarzavery: since we are at it.. i am thinking we should do the change of consuming the latest version of azure-sdk-for-node and complete it. What do you think Yugang?

amarzavery:
LGTM

lib/util/utils.js
yugangw-msft: I thought about that too, but the current decision is to let partner service team do it, right? If we still like do it ourselves, we can do it in the next PR such as partition the utils.js?

PR #1808, opened 2015-05-27
Add a few more packages to list of removing azure-common dependency module

ljhljh235:
Otherwise some files will have a too long absolute path.


yugangw-msft:
It is test flakiness. No way could be caused by the setup change. Merging. Will investigate if the it happens again. 
     cli vm Vm Create:  Create VM with ssh cert and no ssh password, but no ssh endpoint or explicit disabled ssh endpoint should throw error: error: --no-ssh-password and --ssh-cert can only be used with --ssh or --no-ssh-endpoint parameter
error: ENOSPC, no space left on device

PR #1833, opened 2015-06-11
add script to regenerate metadata files during build installer

yugangw-msft:
we need to regenerate because the installer will  pre-compile streamline files so that the file extension needs to be refreshed from '._js' to '.js' in the metadata


amarzavery:
lgtm

PR #1835, opened 2015-06-11
Fix tarball creation script

ljhljh235:


yugangw-msft:
we will submit a good fix next week when we group upgrade the azure-common version to the latest. This change just comments out the invalid scripts, caused by we no longer uses the roll up pacakge

PR #1841, opened 2015-06-15
Fix for change that broke some keyvault commands on 0.9.3

msfcolombo:


azurecla:
Hi __@msfcolombo__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (fcolombo). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/util/utils.js
```diff
@@ -205,7 +205,8 @@ exports.createEventsClient = function (subscription) {
 };
 
 exports.createKeyVaultClient = function (subscription, vaultUri) {
-  var factoryMethod = require('azure-keyvault').createKeyVaultClient;
+  var keyvault = require('azure-keyvault');
+  var factoryMethod = keyvault.createKeyVaultClient;
```
yugangw-msft: The change doesn't appear make any difference from the current code one. Am i missing anything? What was the error log

msfcolombo: The change reintroduces the keyvault variable, which is used below on line 211 (not showing on the diff). The error log is an exception about the undefined variable keyvault.

yugangw-msft: it makes sense now.

PR #1845, opened 2015-06-17
add --remote-subnet option for setting acl-rule

thstarshine:
The --remote-subnet option is missing from commander settings.

### DIFF
--- a/lib/commands/asm/vm._js
+++ b/lib/commands/asm/vm._js
@@ -452,6 +452,7 @@ exports.init = function(cli) {
     .option('-o, --order <id>', $('the ACL rule order'))
     .option('-w, --new-order <id>', $('the new order for the ACL'))
     .option('-a, --action  <name>', $('the new action for the ACL'))
+    .option('-t, --remote-subnet  <subnet>', $('the ACL rule remote subnet'))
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@thstarshine__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

yugangw-msft:
@anuchandy could you take a look

yugangw-msft:
note, i am a little suspicious for whether this change will work for real, as there is no downstream code to use this flag.

thstarshine:
Actually it works well.
The updateEPAclRule() function in lib/commands/asm/vm/vmclient.js has relevant codes to deal with --remote-subnet option. Just need an entry point for it.

```
updateEPAclRule: function(vmName, endpointName, order, options, callback) {
...
    if (options.remoteSubnet) {
...
}
```

anuchandy:
@yugangw-msft , @thstarshine , yes we do have logic present in the 'updateEPAclRule' but we were missing the parameter for the same

yugangw-msft:
Thanks for the contribution

PR #1848, opened 2015-06-18
ensure values returned from prompt is not empty by prompt 

yugangw-msft:
Before, you can press ENTER, and the command continues and fails later; with the change, you will be prompted again, till you input a non-null value.
I could create a new routine for enforced check, but after cross checked all code, i see no reason to accept empty value. If some commands do, then they should not even prompt.

### DIFF
--- a/lib/commands/storage/storage.blob._js
+++ b/lib/commands/storage/storage.blob._js
@@ -551,8 +551,6 @@ exports.init = function(cli) {
     var isOnContainer = !blob;
     container = interaction.promptIfNotGiven($('Container name: '), container, _);
 
-    blob = interaction.promptIfNotGiven($('Blob name: '), blob, _);
-
     if (!options.policy) {
       permissions = interaction.promptIfNotGiven($('Permissions: '), permissions, _);
       if (isOnContainer) {
--- a/lib/util/interaction._js
+++ b/lib/util/interaction._js
@@ -212,12 +212,11 @@ __.extend(Interactor.prototype, {
   },
 
   promptIfNotGiven: function (promptString, currentValue, _) {
-    if (__.isUndefined(currentValue)) {
-      var value = this.prompt(promptString, _);
-      return value;
-    } else {
-      return currentValue;
+    var givenValue = currentValue;
+    while (!givenValue) {
+      givenValue = this.prompt(promptString, _);
     }
+    return givenValue;
   },
 
   choose: function (values, callback) {


lib/commands/storage/storage.blob._js
```diff
@@ -551,8 +551,6 @@ exports.init = function(cli) {
     var isOnContainer = !blob;
     container = interaction.promptIfNotGiven($('Container name: '), container, _);
 
-    blob = interaction.promptIfNotGiven($('Blob name: '), blob, _);
```
yaxia: The original code has a bug when the createBlobSAS() is called by createContainerSAS(), in which case, the blob should be null, but it will trigger 'promptIfNotGiven'.

Removing this call will cause creating blob not working. I've fixed the issue at my side and plan to raise my PR in next sprint. Here is my code for your review: (the current function createContainerSAS() and function createBlobSAS() will all call a new function createSas)

```Javescript
  /**
  * Create shared access signature to the container
  */
  function createContainerSAS(container, permissions, expiry, options, _) {
    createSas(container, null, permissions, expiry, options, true, _);
  }
  
   /**
  * Create shared access signature to the blob
  */
  function createBlobSAS(container, blob, permissions, expiry, options, _) {
    createSas(container, blob, permissions, expiry, options, false, _);
  }

  /**
  * Create shared access signature
  */
  function createSas(container, blob, permissions, expiry, options, isOnContainer, _) {
    var blobService = getBlobServiceClient(options);
    container = interaction.promptIfNotGiven($('Container name: '), container, _);

    if (!isOnContainer) {
      blob = interaction.promptIfNotGiven($('Blob name: '), blob, _);
    }

    if (!options.policy) {
      permissions = interaction.promptIfNotGiven($('Permissions: '), permissions, _);
      if (isOnContainer) {
        StorageUtil.validatePermissions(StorageUtil.AccessType.Container, permissions);
      } else {
        StorageUtil.validatePermissions(StorageUtil.AccessType.Blob, permissions);
      }

      expiry = interaction.promptIfNotGiven($('Expiry: '), expiry, _);
      expiry = validation.parseDateTime(expiry);
    }

    var start;
    if (options.start) {
      start = validation.parseDateTime(options.start);
    }

    var output = { sas: '', url: '' };
    var sharedAccessPolicy = StorageUtil.getSharedAccessPolicy(permissions, start, expiry, null, options.policy);
    var tips;
    if (isOnContainer) {
      tips = util.format($('Creating shared access signature for container %s'), container);
    } else {
      tips = util.format($('Creating shared access signature for blob %s in container %s'), blob, container);
    }
    startProgress(tips);
    try {
      output.sas = blobService.generateSharedAccessSignature(container, blob, sharedAccessPolicy);
      output.url = blobService.getUrl(container, blob, output.sas);
    } finally {
      endProgress();
    }

    cli.interaction.formatOutput(output, function(outputData) {
      logger.data($('Shared Access Signature'), outputData.sas);
      logger.data($('Shared Access URL'), outputData.url);
    });
  }
```

yugangw-msft: Thanks for the detail. Your new code looks better.
One question is regarding where do we strand on the current dev branch. My cross check shows the "createBlobSAS" is only used by "CreateContainerSAS", which always set the 'blob' to 'null'. If that is the case, then removing the prompt is safe because it will be a non-op anyway. You can double check though, the existing logic of "promptIfNotGive" uses the "__.isUndefined(currentValue)", and it will not catch the 'null' value.
Your upcoming PR can always bring back the correct code. 
But let me know if i still miss the context. I just need more data for the right change at right time and let Babu know. This is not a critical change by the way

yaxia: createBlobSas is also called in command "azure storage blob sas create". 
https://github.com/yugangw-msft/azure-sdk-tools-xplat/blob/prompt/lib/commands/storage/storage.blob._js#L278

yugangw-msft: sounds good. let me postpone the fix till the next sprint

yugangw-msft:
Close the PR as it needs refactoring work from the storage team.

PR #1856, opened 2015-06-19
Address a test flakiness in webapp command

yugangw-msft:
Don't access current profile till suite gets set up. Otherwise, the test will hit exception of no existing accounts


yugangw-msft:
reviewed with Naveen

NaveenGoli:
thanks for reviewing

Issue #1831, opened 2015-06-09
ARM: Template Validation Error Messages Are Not Helpful


gabrielhartmann:
All errors in template validation produce errors of this form:

Error: Deployment template validation failed: 'The template resource 'MyUbuntuVM0/DockerExtension' at line '1' and column '7593' is not valid: Template language expression ...

followed by a stack trace.  In any expression of significant complexity these errors don't provide sufficient detail.

1. The line number is always '1' and the column number is meaningless.  I guess this makes sense if I wrote my whole template on one line.

2.  The error messages doesn't give me a clue about what the real problem is.  For example can you find the bug here from the error message above?

"command": "[concat('-server -advertise ', reference(concat(variables(nicName), copyIndex())).ipConfigurations[0].properties.privateIPAddress, ' -join ', reference(concat(variables(nicName), 'master')).ipConfigurations[0].properties.privateIPAddress)]"

The answer is, that the instances of nicName should have single quotes around them.  The unhelpful error messages make template authoring a really slow and painful experience.

yugangw-msft:
Looks like a known issue, #1820 should addresed

yugangw-msft:
FYI, we have published a new version to npm

gabrielhartmann:
@yugangw-msft Are you sure that pull request addresses the error message situation?  Is the cli now producing error messages of the type: "Unexpected non-string token: [nicName] on line 42." ?

yugangw-msft:
@vivsriaus , please take a look. This is a generic validation issue. In a rush to read through, my early comment did not appear correct.

TianoMS:
Checked on PowerShell and we don't have this problem.

vivsriaus:
@TianoMS Can you share the template you used to check if this issues existed in PoSH? I don't see a template being referenced here, in this issue.

TianoMS:
@vivsriaus what I did is get any well-formed template, and make some changes to make it broken (like removing single quotes on a variable name). Try deployment with the template and check the error message.

vivsriaus:
@gabrielhartmann We've made a bunch of changes to error handling in CLI over the past few months. I tried passing in a similar template which had a missing single quote for variable name, and I don't see the generic error (line 1 column x) anymore. Can you please retry?

gabrielhartmann:
[...]

PR #1886, opened 2015-06-26
Handle different data types (int, obj, array, etc) when deploying tem…

vivsriaus:
…plates. Use json.parse for getting template param values


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

yugangw-msft:
please fix jlint error by using single quots

lib/commands/arm/group/groupUtils._js
```diff
@@ -371,6 +371,17 @@ function getTemplateParams(cli, subscription, options, _)
     paramKeys[param] = templateContent.parameters[param].defaultValue;
     if (!templateContent.parameters[param].defaultValue) {
       paramKeys[param] = cli.interaction.promptIfNotGiven(param + ': ', paramKeys[param], _);
+      switch (templateContent.parameters[param].type.toLowerCase()) {
+        case "int":
+          paramKeys[param] = utils.parseInt(paramKeys[param]);
+          break;
+
```
amarzavery: In the template mentioned in the issue, the type of instanceCount is string and not int.
So how will the above code solve the problem?
```
error: Deployment template validation failed: 
'The provided value for the template parameter
 'instanceCount' at line '8' and column '28' is not valid.'.
```
```
"parameters": {
"vmSSName": {
"type": "string"
},
"instanceCount": {
"type": "string",
"metadata": {
"description": "Number of VM instances"
}
},
```

amarzavery:
It will be nice, if you can add a unit test that verifies the issue is fixed.

amarzavery:
@vivsriaus - This seems to be a related issue. Can you fix this as well https://github.com/Azure/azure-xplat-cli/issues/1831 ?

lib/commands/arm/group/groupUtils._js
vivsriaus: The instanceCount type should be an "int". People were using "string" to get around this issue. So, now, if you edit the template and specify instanceCount as "int", the group deployment command should work.

vivsriaus:
@amarzavery I've modified an existing test to include "int" in template file and have captured new records. With regards to improving the error message, we'll take that as a separate PR.

PR #1887, opened 2015-06-27
Fix bug in creating log directory

karlaugsten:
Fix bug where create logging file would fail because mkdir method was not synchronous.

- Log directory would not be created in time on a slower computer, causing the log file creation to throw a 'file does not exist' error.


azurecla:
Hi __@karlaugsten__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

yugangw-msft:
thanks for the contribution.

PR #1890, opened 2015-06-29
Add optional environment to specify resource group.

karlaugsten:
Add optional environment variable AZURE_ARM_TEST_RESOURCE_GROUP in keyvault tests to specify resource group.

- Currently no other tests use the AZURE_ARM_TEST_RESOURCE_GROUP variable.

### DIFF
--- a/test/commands/arm/keyvault/arm.keyvault-tests.js
+++ b/test/commands/arm/keyvault/arm.keyvault-tests.js
@@ -27,7 +27,6 @@ var testUtil = require('../../../util/util');
 var utils = require('../../../../lib/util/utils');
 
 var testPrefix = 'arm-cli-keyvault-tests';
-var testResourceGroup = 'xplatTestVaultRG';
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@karlaugsten__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

yugangw-msft:
@msfcolombo , please take a look. The change looks reasonable, even it touches the recording file, but the change will persist even when you re-record in future

msfcolombo:
@yugangw-msft, the change looks good, I review with @karlaugsten before he submitted. The rationale is to evolve towards allowing full parallel execution of live tests (so names cannot be reused). We manually changed the recordings because it's not using the default value defined in the test itself. We could do re-record but then the delta would be unnecessarily big. Thanks for calling my attention.

yugangw-msft:
sounds good

PR #1904, opened 2015-07-10
Handle v2 version of parameters file during deployment

vivsriaus:
Handle v2 version of parameters file during deployment


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

test/data/arm-deployment-parametersv2.json
```diff
@@ -0,0 +1,17 @@
+{
```
amarzavery: @vivsriaus - if my understanding is correct, then v2 should have top level key as "parameters" like this:
{
  "parameters" : 
  {
    "siteName": {
      "value": "xDeploymentTestSite16754"
    },
    "hostingPlanName": {
      "value": "xDeploymentTestHost24764"
    },
    "siteLocation": {
      "value": "West US"
    },
    "sku": {
      "value": "Free"
    },
    "workerSize": {
      "value": 0
    }
  }
}
```
That will test v2 in true sense.

File: test/commands/arm/group/arm.group.deployment-tests.js [hunk 2]
```diff
@@ -87,13 +87,16 @@ describe('arm', function () {
 
     function setUniqParameterNames(suite, filename) {
       //no need to create unique parameter values in playbackmode
-      var parameters = JSON.parse(fs.readFileSync(filename).toString());
+      var deploymentParameters = JSON.parse(fs.readFileSync(filename).toString());
+      if (deploymentParameters.parameters) {
```
amarzavery: Why would you add this in test code? The command code should handle it for you, right?

markcowl:
@vivsriaus Please respond to feedback

markcowl:
@vivsriaus I believe Amar is correct about V2 parameters files (they also have a schema and contenVersion property).  Please add in a proper test for v2 parameters.  You cna find a schema for V2 parameter files here:  https://github.com/Azure/azure-powershell/blob/dev/src/ResourceManager/Resources/Commands.Resources/Models.ResourceGroups/TemplateFileParameterv2.cs .  Please correct the issues or respond to review feedback.

test/data/arm-deployment-parametersv2.json
vivsriaus: @amarzavery My bad, you're correct. It should include the parameters element. I've fixed this now with my latest commit.

test/commands/arm/group/arm.group.deployment-tests.js
vivsriaus: setUniqParameterNames is called before we execute the command in our tests, to set unique names for parameters. So, this is required to differentiate v2 and v1 parameters file.

ahmetalpbalkan:
@amarzavery I still fail to deploy templates whose parameters file starts like the following:

```
{
  "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
```

and getting error:

```
+ Initializing template configurations and parameters
+ Creating a deployment
error:   The request content was invalid and could not be deserialized: 'Error converting value "http://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#" to type 'Microsoft.WindowsAzure.ResourceStack.Frontdoor.Data.Definitions.DeploymentParameterDefinition'. Path 'properties.parameters.$schema', line 1, position 15622.'.
info:    Error information has been recorded to /Users/alp/.azure/azure.err
error:   group deployment create command failed
```

Issue #1727, opened 2015-04-30
(0.8.16)azure login with  --service-principal  and after token is expired, .azure/accessTokens.json increase to 60M


qingfuwang:
use --service-principal login, and manually change the expired time, to rigger refresh

Then , the size of accesstokean increase to 60M, and azure command would hang 

root@qingfubosh2:/home/azureuser# ls -l /root/.azure/accessTokens.json
-rw------- 1 root root 458565 Apr 30 11:21 /root/.azure/accessTokens.json
root@qingfubosh2:/home/azureuser# azure group list >a.log
root@qingfubosh2:/home/azureuser# ls -l /root/.azure/accessTokens.json
-rw------- 1 root root 462645 Apr 30 11:22 /root/.azure/accessTokens.json
root@qingfubosh2:/home/azureuser# azure group list >a.log
root@qingfubosh2:/home/azureuser# ls -l /root/.azure/accessTokens.json
-rw------- 1 root root 466725 Apr 30 11:22 /root/.azure/accessTokens.json
root@qingfubosh2:/home/azureuser# azure group list >a.log
root@qingfubosh2:/home/azureuser# ls -l /root/.azure/accessTokens.json
-rw------- 1 root root 470805 Apr 30 11:22 /root/.azure/accessTokens.json
root@qingfubosh2:/home/azureuser# azure group list >a.log
root@qingfubosh2:/home/azureuser# ls -l /root/.azure/accessTokens.json
-rw------- 1 root root 945445 Apr 30 11:22 /root/.azure/accessTokens.json
root@qingfubosh2:/home/azureuser# azure group list >a.log
root@qingfubosh2:/home/azureuser# azure group list >a.log
root@qingfubosh2:/home/azureuser# azure group list >a.log
root@qingfubosh2:/home/azureuser# azure group list >a.log
root@qingfubosh2:/home/azureuser# azure group list >a.log
root@qingfubosh2:/home/azureuser# ls -l /root/.azure/accessTokens.json
-rw------- 1 root root 15249925 Apr 30 11:23 /root/.azure/accessTokens.json
root@qingfubosh2:/home/azureuser# azure group list >a.log
root@qingfubosh2:/home/azureuser# azure group list >a.log
^Croot@qingfubosh2:/home/azureuser#  ls -l /root/.azure/accessTokens.json
-rw------- 1 root root 61008485 Apr 30 11:28 /root/.azure/accessTokens.json

amarzavery:
[...]

PR #1927, opened 2015-07-16
Ensure a service principal gets logged out cleanly(all cached tokens get cleaned up)

yugangw-msft:
Fix https://github.com/Azure/azure-xplat-cli/issues/1727
It appears when we introduced the service principal, we never wired up the logging out. 
This is not a linux specific bug; rather it gets more visible there because the token cache is file based.


lib/commands/login._js
```diff
@@ -113,11 +113,12 @@ exports.init = function (cli) {
   cli.command('logout [username]')
     .description($('Log out from Azure subscription using Active Directory. Currently, the user can log out only via Microsoft organizational account'))
     .option('-u --username <username>', $('Required. User name used to log out from Azure Active Directory.'))
+    .option('--service-principal', $('If given, log out the service principal'))
```
amarzavery: It is not necessary to add --service-principal flag.

Issue #1955, opened 2015-07-24
Completion does not autocomplete filenames


ahmetalpbalkan:
I just added `azure --completion` script to my `~/.zshrc` and tried to deploy a template using command. I typed:

    azure group deployment create -n "myName" -f 

then I hit `<tab>` `<tab>` and nothing happened. So I had to Ctrl+C and run a `ls` to see what the file was actually named. I think the completion should somehow detect `-f` (and likewise `-e` in this command) is a filename input and do a glob in the PWD.

devigned:
That would be really helpful.

yugangw-msft:
I will take a look.

ahmetalpbalkan:
@yugangw-msft can you please take a look here? This does not fix the bug entirely

0.9.10:

    $ azure vm extension set [bla bla] --public-config-path <tab><tab> <-- does not work

can you please reopen?

yugangw-msft:
@ahmetalpbalkan, my PR introduced a new file related flag so auto-complete can recognize as a file related option and do a "ls". At that time I did convert some existing options though, but obviously i could not know all of them. 
Re-think now, maybe i don't even need to introduce a file flag; rather i should just do a "ls", even for a not file related option. Not that experienced on non-windows platform, but 2 reasons:
1. In mac I tried out several well-known commands, looks like a built-in behavior that 2 tabs means "ls".
2. We will have no need to convert file related options in azure-cli, which might have lot.

PR #1963, opened 2015-07-29
More auto-completions for commands

yugangw-msft:
With this change, the cli's auto-completion will be much more completed that at every level it gives out the list.
1. Complete at option level
2. Complete for file related option with files under current working folder. This Fixes https://github.com/Azure/azure-xplat-cli/issues/1955. Note i will follow up with service teams to get all file options updated.


devigned:
Awesome!!

ahmetalpbalkan:
this is cool

PR #1969, opened 2015-07-31
AzureRT PR S52 #1: Add Tags for VM Create

huangpf:


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

huangpf:
@yugangw-msft @amarzavery 
Why are there some extra characters generated in the nock data?
```
}]]
```

lib/commands/arm/vm/vm._js
```diff
@@ -96,6 +96,8 @@ exports.init = function (cli) {
     .option('-Y, --data-disk-existing', $('Will use existing VHD if specified. Don\'t specify this param if you are creating data disk from a new VHD.'))
     .option('-C, --custom-data <custom-data-file>', $('CustomData file'))
     .option('-s, --subscription <subscription>', $('the subscription identifier'))
+    .option('-T, --tags <tags>', $('Tags to set to the resource group. Can be multiple. ' +
```
yugangw-msft: i would suggest drop "-T" and keep "--tags". "-t" is already used for a very different thing. Casing difference only is not explicit to get it right,

yugangw-msft: Also since you are adding new option, please run "node bin/azure --gen arm" to refresh the command metadata file

yugangw-msft:
LGTM overall

huangpf:
@yugangw-msft Thanks Yugang. I had addressed your comments and updated the PR. Could you please help take a look again? Thanks.

lib/commands/arm/vm/vm._js
```diff
@@ -96,6 +96,8 @@ exports.init = function (cli) {
     .option('-Y, --data-disk-existing', $('Will use existing VHD if specified. Don\'t specify this param if you are creating data disk from a new VHD.'))
     .option('-C, --custom-data <custom-data-file>', $('CustomData file'))
     .option('-s, --subscription <subscription>', $('the subscription identifier'))
+    .option('-tags, --tags <tags>', $('Tags to set to the resource group. Can be multiple. ' +
```
yugangw-msft: Sorry what i meant was you just use "--tags <tags>" without a short option. Ideally we should use "-t, --tags", like other commands, but yours has already used "-t" for other very different purpose

huangpf: I thought you meant only ```--tags``` as well, but is it fine if I keep both ```-tags``` and ```--tags```? Thanks.

yugangw-msft: Then it will be inconsistent with other commands, as the short option implies a "-" and a single character

lib/commands/arm/vm/vm._js
```diff
@@ -96,6 +96,8 @@ exports.init = function (cli) {
     .option('-Y, --data-disk-existing', $('Will use existing VHD if specified. Don\'t specify this param if you are creating data disk from a new VHD.'))
     .option('-C, --custom-data <custom-data-file>', $('CustomData file'))
     .option('-s, --subscription <subscription>', $('the subscription identifier'))
+    .option('--tags <tags>', $('Tags to set to the virtual machine. Can be multiple. ' +
+            'In the format of \'name=value\'. Name is required and value is optional. For example, -t tag1=value1;tag2'))
```
yugangw-msft: the help information is till not not right. Please update it to be " --tag tag1=value1;tag2"

huangpf: Thanks. Good Catch~

huangpf:
retest this please

PR #1972, opened 2015-08-03
update cli so to depend on only one version azure-common

yugangw-msft:
All node lib have been updated to use sliding version of azure-common, and cli now leverages it to handle some existing issues:
1. This resolves the long file path issue on building windows msi
2. Small file size for both windows and mac installer
3. No need to have explict (and not maintabable) script to prune azure-common under node-modules folder


test/commands/cli.network-tests.js
```diff
@@ -29,8 +32,10 @@ var requiredEnvironment = [{
 var testSite;
 
 describe('cli', function() {
-  describe('network', function() {
-    var networkconfig = 'netconfig.json';
+  describe('network', function () {
+    //put the json file under same folder of the test file; 
+    //rather under repo root.
+    var networkconfig = path.join(__dirname, 'netconfig.json');
```
amarzavery: How about putting it under tests/output? In this way it wont interfere while committing tests as the entire tests/output folder is marked in .gitignore file.

PR #1991, opened 2015-08-06
Authorization: Changes for Create and Update of Role definitions

namratab:
This change adds the 'create' and 'set' options for role definitions (custom roles). 
Delete will be added in a separate PR at which point the tests will be updated to clean-up


azurecla:
Hi __@namratab__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (nbaliga). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/commands/arm/role/role.definition._js
```diff
@@ -54,33 +56,117 @@ exports.init = function (cli) {
...
 
-    cli.interaction.formatOutput(roles, function (data) {
-      if (data.length === 0) {
-        log.info($('No role definition found'));
-      } else {
-        log.table(data, displayARoleDefinition);
+  role.command('create [inputfile] [roledefinition]')
+    .description($('create a new role definition'))
+    .option('-f --inputfile <inputfile>', $('File name containing a single role definition.'))
+    .option('-r --roledefinition <roledefinition>', $('A JSON-formatted string containing the role definition'))
```
amarzavery: How would the end user know what to provide. Please provide an example here. 

'A JSON-formatted string containing the role definition. For example - "{"name": "custome_role", blah blah"}" '

lib/commands/arm/role/role.definition._js
```diff
@@ -54,33 +56,117 @@ exports.init = function (cli) {
...
+        if (role) {
+          roleUtils.showRoleDefinition(role, log);
+        }
+      });
+    });
+
+  role.command('set [inputfile] [roledefinition]')
+    .description($('update an existing role definition'))
+    .option('-f --inputfile <inputfile>', $('File name containing a single role definition.'))
+    .option('-r --roledefinition <roledefinition>', $('A JSON-formatted string containing the role definition'))
```
amarzavery: Please provide an example.

lib/commands/arm/role/roleUtils._js
```diff
@@ -0,0 +1,172 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+var underscore = require('underscore');
```
amarzavery: This is fine as well. However, in js world, people usually declare it as follows:
```js
var __ = require('underscore');
```

lib/commands/arm/role/roleUtils._js
```diff
@@ -0,0 +1,172 @@
...
+    assignableScopes: roleToUpdateAssignableScopes,
+    description: roleToUpdateDescription,
+    permissions: [
+      {
+        actions: roleToUpdateActions,
+        notActions: roleToUpdateNotActions
+      }
+    ],
+    roleName: roleToUpdateRoleName,
+    type: 'CustomRole'
```
amarzavery: you are specifying CustomRole over here in pascal case, where as in showRoleDefinition the format is all lowercase
var CUSTOM_ROLE_TYPE = 'customrole'; 

How about making this as a constant in this file https://github.com/Azure/azure-xplat-cli/blob/dev/lib/util/constants.js and then use it every where. Shubham is also using this in role list. In this way one value can be used every where.

test/commands/arm/role/arm.role-tests.js
```diff
@@ -224,6 +227,94 @@ describe('arm', function () {
...
+        suite.execute('role create -f %s --json', filePath, function (result) {
+          result.exitStatus.should.equal(0);
+          var createdRole = JSON.parse(result.text);
+          createdRole.roleDefinition.properties.roleName.should.equal("CustomRole Test");
+          createdRole.roleDefinition.properties.assignableScopes.length.should.be.above(0);
+          createdRole.roleDefinition.properties.assignableScopes[0].should.equal("/subscriptions/4004a9fd-d58e-48dc-aeb2-4a4aec58606f");
+          createdRole.roleDefinition.properties.permissions.length.should.be.above(0);
+          createdRole.roleDefinition.properties.permissions[0].actions.length.should.be.above(0);
+            
+          // TODO: Clean up role after delete is implemented
```
amarzavery: Can you implement the cleanup as well?
Or you can collect all the role definition names in an array and in the "after" section iterate over the array to delete all the roles

test/commands/arm/role/arm.role-tests.js
```diff
@@ -224,6 +227,94 @@ describe('arm', function () {
...
+          roleToUpdate.name = "Updated Role Name";
+          roleToUpdate.description = "Updated Role Description";
+
+          suite.execute('role set -r %s --json', JSON.stringify(roleToUpdate), function(updatedResult) {
+            updatedResult.exitStatus.should.equal(0);
+            var updatedRole = JSON.parse(updatedResult.text);
+            updatedRole.roleDefinition.properties.roleName.should.equal("Updated Role Name");
+            updatedRole.roleDefinition.properties.description.should.equal("Updated Role Description");
+            updatedRole.roleDefinition.properties.assignableScopes.length.should.be.above(0);
+            updatedRole.roleDefinition.properties.assignableScopes[0].should.equal("/subscriptions/4004a9fd-d58e-48dc-aeb2-4a4aec58606f");
```
amarzavery: This test will fail when we run live tests at the end of every release as our subscription id would be different.
Test it like this: https://github.com/Azure/azure-xplat-cli/blob/dev/test/commands/arm/role/arm.role-tests.js#L249

test/commands/arm/role/arm.role-tests.js
```diff
@@ -224,6 +227,94 @@ describe('arm', function () {
...
+            done();
+          });
+        });
+      });
+
+      it('fails to update non-existent role', function(done) {
+        var filePath = path.join(__dirname, '../../../data/CustomRoleDefValid.json');
+        var roleToUpdate = JSON.parse(fs.readFileSync(filePath));
+        // random GUID for role id
+        roleToUpdate.id = "/subscriptions/4004a9fd-d58e-48dc-aeb2-4a4aec58606f/providers/Microsoft.Authorization/roleDefinitions/43367f6e-e106-480d-a448-2a393ea5eb21";
```
amarzavery: This test will not pass when we run live tests during end of release activities.
The roleToUpdate.id has a subscription id that we do not have access to. Hence we will fail at authentication level.
I would suggest to create the id dynamically by getting the current subscription id and then having some random guid for role definition

lib/commands/arm/role/role.definition._js
namratab: Fixed

lib/commands/arm/role/role.definition._js
namratab: Added

lib/commands/arm/role/roleUtils._js
namratab: Made the change. Thanks

lib/commands/arm/role/roleUtils._js
namratab: Instead of updating the global constants file, I've added a new constants file specific to rbac

test/commands/arm/role/arm.role-tests.js
namratab: As discussed offline, Shubham will be implementing the delete in a day or so and he will update these tests to clean up

test/commands/arm/role/arm.role-tests.js
namratab: I am still checking for a particular id, but using the current subscription always instead of hard coding it

test/commands/arm/role/arm.role-tests.js
namratab: Thanks for pointing this out, I've made the change

PR #2001, opened 2015-08-07
Support storage service version 2015-02

yaxia:
1. Update azure-storage to 0.5.0 which supports Azure storage service version 2015-02-21
2. Support append blob
3. Supports share quota and share usage
4. New commands for shared access signatures for shares and files
  *  azure storage share sas create [options] [share] [permissions] [expiry] 
  *  azure storage file sas create [options] [share] [path] [permissions] [expiry]
5. New commands for share ACL
  * azure storage share policy create [options] [share] [name]
  * azure storage share policy show [options] [share] [name]
  * azure storage share policy list [options] [share]
  * azure  storage share policy set [options] [share] [name]
  * azure storage share policy delete [options] [share] [name] 
6. New commands for file async copy
  * azure storage file copy start [options] [sourceUri] [destShare]
  * azure storage file copy show [options] [share] [path]
  * azure storage file copy stop [options] [share] [path] [copyid]
7. New commands for CORS (Cross-Origin Resource Sharing)
  * azure storage cors set [options]
  * azure storage cors show [options]
  * azure storage cors delete [options]


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/commands/storage/storage.blob._js
```diff
@@ -229,17 +225,18 @@ exports.init = function(cli) {
...
+    .description($('Start to copy the resource to the specified storage blob which completes asynchronously'))
     .option('--source-sas <sourceSas>', $('the shared access signature of the source storage'))
-    .option('--source-container <sourceContainer>', $('the source storage container name'))
-    .option('--source-blob <sourceBlob>', $('the source storage blob name'))
-    .option('--source-uri <sourceUri>', $('the source storage blob absolute uri'))
+    .option('--source-uri <sourceUri>', $('the source storage blob or file absolute uri'))
+    .option('--source-container <sourceContainer>', $('the source storage container name when copies a blob to a blob'))
+    .option('--source-blob <sourceBlob>', $('the source storage blob name when copies a blob to a blob'))
+    .option('--source-share <sourceShare>', $('the source storage share name when copies a file to a blob'))
+    .option('--source-path <sourcePath>', $('the source storage file path when copies a file to a blob'))
```
yugangw-msft: if this is for a local file, i suggest you replace with the "fileRelatedOption" which the benefit that the auto complete will be able to display files under cwd.

lib/commands/storage/storage.file._js
```diff
@@ -270,6 +418,36 @@ exports.init = function(cli) {
...
+    startProgress(tips);
+    try {
+      var getPropertiesOperation = getStorageFileOperation(fileService, 'getShareProperties');
+      var storageOptions = getStorageFileOperationDefaultOption();
+      properties = performStorageOperation(getPropertiesOperation, _, share, storageOptions);
+
+      var setPropertiesOperation = getStorageFileOperation(fileService, 'setShareProperties');
+      if (options.quota) {
+        if (!validation.isInt(options.quota)) {
+          throw new Error(util.format($('The share quota must be an integer in GB')));
```
yugangw-msft: i would suggest do this validation before real operations starts, so move this to the beginning of method.

lib/commands/storage/storage.service._js
```diff
@@ -263,27 +299,147 @@ exports.init = function(cli) {
...
+  */
+  function deleteCORS(options, _) {
+    var types = getServiceTypes(options, true, true);
+    var client = getServiceClient(types[0], options);
+    var getOperation = StorageUtil.getStorageOperation(client, types[0], 'getServiceProperties');
+    var setOperation = StorageUtil.getStorageOperation(client, types[0], 'setServiceProperties');
+    var force = !!options.quiet;
+
+    if (force !== true) {
+      force = cli.interaction.confirm($('Do you want to delete the CORS rules?'), _);
```
yugangw-msft: need to add "[y/n]" to the message?

lib/commands/storage/storage.service._js
```diff
@@ -494,4 +639,58 @@ exports.init = function(cli) {
...
+  }
+
+  /**
+  * @ignore
+  * Validates CORS rule maximum age in seconds
+  * @param {string} [methods] the allowed methods in the rule
+  */
+  function parseAndValidateCORSRuleMaxAge(age) {
+    if (age && !validation.isInt(age)) {
+      throw new Error($('The --max-age should be an integer'));
```
yugangw-msft: should we complain about the "age", rather the "--max-age", which is not in the context?

lib/util/storage.util._js
```diff
@@ -1133,27 +1212,485 @@ function getStorageServiceSettings(connection) {
...
+  
+  // Check whether it is logged in, otherwise read from the default environment.
+  if (__.keys(profile.current.subscriptions).length > 0) {
+    uri = utils.createStorageClient(profile.current.getSubscription()).baseUri;
+  } else {
+    uri = profile.current.environments.AzureCloud.managementEndpointUrl;
+  }
+  
+  var delimiter = uri.indexOf('/');
+  if (delimiter === -1) {
```
yugangw-msft: In case it helps, have you thought about leveraging "url.parse()"?

lib/util/storage.util._js
```diff
@@ -1133,27 +1212,485 @@ function getStorageServiceSettings(connection) {
...
+
+/**
+* Start async copy
+*/
+StorageUtil.startAsyncCopy = function (startCopyParams, sourceUri, destContainerOrShare, options, _) {
+  var isCopyToBlob = isCopyDestinationBlob(startCopyParams.type);
+  
+  // Make up the source blob or file Uri
+  var source = getSourceUri(startCopyParams.copyType, options, _);
+  var sourceContainerOrShare = options.isSourceBlobDefined ? options.sourceContainer : (options.isSourceFileDefined ? options.sourceShare : undefined);
```
yugangw-msft: please break the line into 2.

lib/util/storage.util._js
```diff
@@ -1133,27 +1212,485 @@ function getStorageServiceSettings(connection) {
...
+    if (!sourceResource.blobOrFilePath) {
+      if (options.isSourceBlobDefined) {
+        throw new Error($('The source blob is invalid'));
+      } else if (options.isSourceFileDefined) {
+        throw new Error($('The source file is invalid'));
+      }
+    } else {
+      // Parse the blob name or file path from URI when sourceUri is specified with SAS token.
+      var index = sourceResource.blobOrFilePath.lastIndexOf('?');
+      if (index != -1) {
```
yugangw-msft: !==

lib/util/storage.util._js
```diff
@@ -1133,27 +1212,485 @@ function getStorageServiceSettings(connection) {
...
+  sourceResource.blobOrFilePath = sourceResource.blobName;
+  
+  if (!sourceContainerOrShare) {
+    sourceContainerOrShare = sourceResource.containerOrShare;
+  }  
+  
+  if (!sourceBlobOrFile) {
+    if (!sourceResource.blobOrFilePath) {
+      if (options.isSourceBlobDefined) {
+        throw new Error($('The source blob is invalid'));
```
yugangw-msft: [Up to you] would be great to have the error tell users how to set the source blob correctly, say, they should at least provide "sourceBlobOrFile" or "blobOrFilePath", etc

lib/util/storage.util._js
```diff
@@ -1133,27 +1212,485 @@ function getStorageServiceSettings(connection) {
...
+  // Make sure the read access to the source blob or file
+  var sasToken = getSourceSASToken(startCopyParams, sourceContainerOrShare, sourceBlobOrFile, embeddedSas, options, _);
+  source += (sasToken === undefined ? '' : ('?' + sasToken));
+
+  // Check destination information
+  var destBlobOrFile;
+  if (isCopyToBlob) {
+    destBlobOrFile = options.destBlob ? options.destBlob : sourceBlobOrFile;
+  } else {
+
```
yugangw-msft: you probably do not need this line

lib/util/storage.util._js
```diff
@@ -1133,27 +1212,485 @@ function getStorageServiceSettings(connection) {
...
+  // Start copy operation
+  var destOption = getDestAccountOptions(options);
+  var destServiceClient = StorageUtil.getServiceClient(isCopyToBlob ? StorageUtil.getBlobService : StorageUtil.getFileService, options);
+  var startOperation = startCopyParams.getStartOperation(destServiceClient);
+  var tips = util.format(isCopyToBlob ? $('Start copying blob %s') : $('Start copying file %s'), source);
+
+  var copyOp;
+  var retry;
+  StorageUtil.startProgress(tips);
+  while (retry === true || retry === undefined) {
```
yugangw-msft: could you simplify the condition by not letting "undefined" stay as a valid value? the big loop below doesn't really produce a "undefined" value. When you define the "variable", you can just set to "true".

lib/util/storage.util._js
```diff
@@ -1133,27 +1212,485 @@ function getStorageServiceSettings(connection) {
...
+StorageUtil.showAsyncCopy = function (showCopyParams, containerOrShare, blobOrFilePath, options, _) {
+  var isCopyToBlob = isCopyDestinationBlob(showCopyParams.type);
+
+  containerOrShare = cli.interaction.promptIfNotGiven(isCopyToBlob ? $('Destination container: ') : $('Destination share: '), containerOrShare, _);
+  blobOrFilePath = cli.interaction.promptIfNotGiven(isCopyToBlob ? $('Destination blob: ') : $('Destination file path: '), blobOrFilePath, _);
+
+  var destOption = getDestAccountOptions(options);
+  var blobOrFileProps = showCopyParams.getProperties(containerOrShare, blobOrFilePath, destOption, _);
+
+  if (blobOrFileProps.copyId === undefined) {
```
yugangw-msft: need to handle null, 0, '', etc?

lib/util/storage.util._js
```diff
@@ -1133,27 +1212,485 @@ function getStorageServiceSettings(connection) {
...
+  var source;
+  if (!isSourceDefined) {
+    source = cli.interaction.promptIfNotGiven($('Source URI: '), source, _);
+  } else {
+    var host;
+    var settings = getStorageServiceSettings(options);
+    
+    if (options.isSourceBlobDefined) {
+      host = settings._blobEndpoint.primaryHost;
+      if (host === undefined) {
```
yugangw-msft: i assume you know the other falsey (like null) value won't be in the context. If yes, then code here is just fine.

yugangw-msft: Please cross-check other code which also specifically checks for "undefined" but not other falsey values

lib/util/storage.util._js
yugangw-msft: please do the same on other long lines. Suggest bring down to 100 characters max

yugangw-msft:
LGTM, overall. Please address review feedback

lib/commands/storage/storage.blob._js
yaxia: It is not for a local file. It is an Azure storage file in cloud.

lib/commands/storage/storage.file._js
yaxia: Resolved.

lib/commands/storage/storage.service._js
yaxia: Resolved.

lib/commands/storage/storage.service._js
yaxia: Resolved.

lib/util/storage.util._js
yaxia: Thanks for the advice. I will put in into next sprint to reserve more time to test the change.

lib/util/storage.util._js
yaxia: Resolved.

lib/util/storage.util._js
yaxia: Resolved.

lib/util/storage.util._js
yaxia: When "options.isSourceBlobDefined" or "options.isSourceFileDefined" is true, but the code still goes here, it means the source is provided but it contains error, e.g. bad format, wrong parameter set, etc. We cannot identify the exact reason here.

lib/util/storage.util._js
yaxia: Resolved.

lib/util/storage.util._js
yaxia: Retry === undefined is the initial value. At the first round, "retry" equals "true" or "undefined" will lead to different behavior, though "undefined" will not be produced in the loop. I understand it could be confusing. I will try to refactor in next sprint.

lib/util/storage.util._js
yaxia: It is returned from the storage SDK, when there is no copy action on the specific blob/file, the field is undefined.

PR #2069, opened 2015-08-28
AzureRT S53 - Set/Get Diagnostics Profile & Log

huangpf:


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/commands/arm/vm/vm._js
```diff
@@ -98,6 +98,10 @@ exports.init = function (cli) {
     .option('-s, --subscription <subscription>', $('the subscription identifier'))
     .option('--tags <tags>', $('Tags to set to the virtual machine. Can be multiple. ' +
             'In the format of \'name=value\'. Name is required and value is optional. For example, --tags tag1=value1;tag2'))
+    .option('--enable-boot-diagnostics <enable-boot-diagnostics>', $('To enable or disable the boot diagnostics for VM. ' +
```
amarzavery: Why do we want the user to provide a value like true or false?
How about the following experience:
- By default the boot-diagnostics is disabled. If the user wants it then he should just specify the switch --enable-boot-diagnostics.

Look at the following example in rediscache create command. 
https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/rediscache/rediscache._js#L60

lib/commands/arm/vm/vmClient._js
```diff
@@ -899,6 +994,44 @@ __.extend(VMClient.prototype, {
...
+        accountName : null,
+        containerName : null,
+        blobName : null
+    };
+
+    var parsedResult = url.parse(serialConsoleLogBlobUri, true);
+    var hostName = parsedResult.host || '';
+    var hostSegments = hostName.split('.');
+    var containerAndBlobPath = parsedResult.path || '';
+    if (containerAndBlobPath.length > 0 && containerAndBlobPath.charAt(0) == '/') {
```
amarzavery: use === for comparison in javascript
http://stackoverflow.com/questions/359494/does-it-matter-which-equals-operator-vs-i-use-in-JavaScript-comparisons

lib/commands/arm/vm/vmClient._js
```diff
@@ -362,6 +377,86 @@ __.extend(VMClient.prototype, {
...
+        if (instanceView && instanceView.bootDiagnostics && instanceView.bootDiagnostics.serialConsoleLogBlobUri)
+        {
+          var serialConsoleLogBlobUri = instanceView.bootDiagnostics.serialConsoleLogBlobUri;
+          this.cli.output.info(util.format($('Serial Console Log Blob Uri:\n%s'), serialConsoleLogBlobUri));
+
+          //Init StorageUtil
+          StorageUtil.init(this.cli);
+
+          var result = this._getStorageAccountContainerAndBlobFromUri(serialConsoleLogBlobUri);
+          this.cli.output.info(util.format($('serialConsoleLogBlobUri = "%s"'), serialConsoleLogBlobUri));
```
amarzavery: Why would you want to the output the same info in two different formats:
- **Serial Console Log Blob Uri**
-  **serialConsoleLogBlobUri**

amarzavery:
@huangpf - Everything else looks good. When do you plan to move the compute node sdk from lib/commands/arm/armsdk to the actual package "azure-arm-compute" package that resides in node sdk?

lib/commands/arm/vm/vmClient._js
huangpf: Duplicate code - I would remove the second one...

huangpf:
Thanks @amarzavery. I would address your comments and create an item to follow up your request of having the actual package of compute by the next PR.

lib/commands/arm/armsdk/compute.js
```diff
@@ -20,16 +20,16 @@ var ComputeManagementClient = require('./computeManagementClient');
 exports.ComputeManagementClient = ComputeManagementClient;
```
yugangw-msft: please remove this file and rather reference the latest compute node sdk lib

yugangw-msft: and also remove the "/computeManagementClient.js" below

lib/commands/arm/vm/vmClient._js
```diff
@@ -30,8 +30,20 @@ var VirtualMachine = require('./virtualMachine');
 var AvailabilitySet = require('./../availabilityset/availabilitySet');
 var VMImage = require('./vmImage');
 
+var StorageUtil = require('../../../util/storage.util');
+var url = require('url');
```
yugangw-msft: please move the url reference to top. The sequence of 'require' is
1. Node build-in lib
2. 3rd path lib
3. Ours
For each category, sort by alphabetic

lib/commands/arm/vm/vm._js
```diff
@@ -318,10 +336,13 @@ exports.init = function (cli) {
             'In the format of \'name=value\'. Name is required and value is optional. For example, -t tag1=value1;tag2'))
     .option('-T, --no-tags', $('Remove all tags'))
     .option('-s, --subscription <subscription>', $('the subscription identifier'))
+    .option('--disable-boot-diagnostics', $('The switch parameter to disable the boot diagnostics for VM.'))
+    .option('--enable-boot-diagnostics', $('The switch parameter to enable the boot diagnostics for VM.'))
```
amarzavery: The vm create command uses one switch --enable-boot-diagnostics whereas the vm set command uses two switches
--enable-boot-diagnostics and --disable-boot-diagnostics

- Why do you want to provide two separate switches for a task that can be accomplished in one switch?

PR #2077, opened 2015-09-02
fix network vnet delete when there is only one network

freedev:
the command network vnet delete network-name does not work when there is only one existing network

### DIFF
--- a/lib/commands/asm/network/virtualNetwork._js
+++ b/lib/commands/asm/network/virtualNetwork._js
@@ -177,7 +177,7 @@ __.extend(VirtualNetwork.prototype, {
     var vnetConfig = networkConfig.VirtualNetworkConfiguration;
 
     var index = utils.indexOfCaseIgnore(vnetConfig.VirtualNetworkSites, {Name: vnetName});
-    if (index) {
+    if (index != null) {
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@freedev__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

romangromov:
@freedev use `!==` instead

yugangw-msft:
The return value of  "indexOfCaseIgnore" looks rather odd that returns "null" if not found. Should have been '-1'

freedev:
I have done an extensive search on the entire code base, there are 32 occurrences of indexOfCaseIgnore, and almost in every case there is a check for null.
Please forgive me if this is a dumb question: I agree that null is an odd return value for a method like "indexOfCaseIgnore", but even if we change the code everywhere from null to -1, shouldn't all developers that right now use this method agree this change of behaviour?

freedev:
@yugangw-msft please take a look at:
- lib/commands/asm/network/dnsServer._js [line 126]
This could be a latent bug.

yugangw-msft:
@freedev yes that is bug.
Your PR has been merged. Feel free to work with Roman and submit a new PR to improve it

romangromov:
@freedev @yugangw-msft I'm looking into this issue, will share update on this, thank you all guys

PR #2080, opened 2015-09-04
Utils: indexOfCaseIgnore function improved

romangromov:
This PR contains minor fix for `utils.indexOfCaseIgnore()` function, now it return `-1` instead of `null` when item not found. Also fixed all usages of this function.
Related issue fixed, reported here https://github.com/Azure/azure-xplat-cli/pull/2077

### DIFF
--- a/lib/util/utils.js
+++ b/lib/util/utils.js
@@ -1117,7 +1117,7 @@ exports.findFirstCaseIgnore = function (list, attributes) {
 };
 
 /**
- * Looks through the list and returns the index of element that matches attributes,
+ * Looks through the list and returns the index of element that matches attributes or -1 if it is not present,
  * case-insensitive.
  */
 exports.indexOfCaseIgnore = function (list, attributes) {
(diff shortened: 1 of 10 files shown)


azurecla:
Hi __@romangromov__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

lib/util/utils.js
```diff
@@ -1129,7 +1129,7 @@ exports.indexOfCaseIgnore = function (list, attributes) {
     var match = exports.isAttributesMatched(item, attributes);
     if (match) return i;
   }
-  return null;
+  return -1;
```
yugangw-msft: at line 1125, should we return -1 as well?
Also as pointed out in the  #2077, should we address the "lib/commands/asm/network/dnsServer._js [line 126]" as well?

romangromov: as for `utils.js line 1125` - yes, sure, fixed
as for `dnsServer._js line 126` - no action required, we are good here. I've check

Issue #2102, opened 2015-09-15
Fresh installation of latest version of Azure-CLI breaks


gkgupta:
I was doing setup on a new machine and installed latest azure-cli:
sudo npm install azure-cli -g

And it is giving error while executing:

```
azure storage blob download -a <STORAGE> -k <KEY> <STORAGE_CONTAINER> <BLOB_NAME>
info:    Executing command storage blob download
error:   Cannot read property 'substr' of null
info:    Error information has been recorded to /root/.azure/azure.err
error:   storage blob download command failed
```

gkgupta:
Error details

```
Tue Sep 15 2015 05:44:01 GMT+0000 (UTC):
{ [TypeError: Cannot read property 'substr' of null] __frame: undefined, rawStack: [Getter] }
TypeError: Cannot read property 'substr' of null
    at Object.exports.stringStartsWith (/usr/lib/node_modules/azure-cli/lib/util/utilsCore.js:67:16)
    at Object.exports.isIP (/usr/lib/node_modules/azure-cli/lib/util/validation.js:38:17)
    at getStorageServiceEndpointSuffix (/usr/lib/node_modules/azure-cli/lib/util/storage.util._js:1139:18)
    at getFullConnectionString (/usr/lib/node_modules/azure-cli/lib/util/storage.util._js:1148:24)
    at adjustConnectionStringWithEnvironment (/usr/lib/node_modules/azure-cli/lib/util/storage.util._js:1196:10)
    at getStorageServiceSettings (/usr/lib/node_modules/azure-cli/lib/util/storage.util._js:1088:24)
    at StorageUtil.getBlobService (/usr/lib/node_modules/azure-cli/lib/util/storage.util._js:139:25)
    at Object.StorageUtil.getServiceClient (/usr/lib/node_modules/azure-cli/lib/util/storage.util._js:522:27)
    at getBlobServiceClient (/usr/lib/node_modules/azure-cli/lib/commands/storage/storage.blob._js:67:37)
    at __$listAzureBlob (/usr/lib/node_modules/azure-cli/lib/commands/storage/storage.blob._js:813:21)
```

ChoOo7:
I have the same issue.
Do you have any workaround ?

yugangw-msft:
@yaxia please investigate. #2103 is the same issue

ChoOo7:
Tested with node 0.10.25 and 0.12.0
from a fresh 
[...]

Issue #2103, opened 2015-09-15
0.9.9: Cannot call method 'substr' of null


AbelHu:
I installed 0.9.9 in Ubuntu 14.04.2 LTS. I always get an error 'Cannot call method 'substr' of null' when using 'azure storage blob' commands. 

command:
azure storage blob show --container CONTAINER-NAME -a STORAGE-ACCOUNT -k STORAGE-KEY

Error:
info:    Executing command storage blob show
error:   Cannot call method 'substr' of null
info:    Error information has been recorded to /home/MYACCOUNT/.azure/azure.err
error:   storage blob show command failed

.azure/azure.err:
Tue Sep 15 2015 07:34:48 GMT+0000 (UTC):
{ [TypeError: Cannot call method 'substr' of null]
  stack: [Getter/Setter],
  __frame:
   { name: 'showAzureBlob',
     line: 749,
     file: '/usr/local/lib/node_modules/azure-cli/lib/commands/storage/storage.blob._js',
     prev: undefined,
     calls: 1,
     active: false,
     offset: 1,
     col: 15 },
  rawStack: [Getter] }
TypeError: Cannot call method 'substr' of null
  <<< async stack >>>
    at Object.exports.stringStartsWith (/usr/local/lib/node_modules/azure-cli/lib/util/utilsCore.js:67:17)
    at Object.exports.isIP (/usr/local/lib/node_modules/azure-cli/lib/util/validation.js:38:17)
    at getStorageServiceEndpointSuffix (/usr/local/lib/node_modules/azure-cli/lib/util/storage.util._js:1139:18)
    at getFullConnectionString (/usr/local/lib/node_modules/azure-cli/lib/util/storage.util._js:1148:24)
    at adjustConnectionStringWithEnvironment (/usr/local/lib/node_modules/azure-cli/lib/util/storage.util._js:1196:10)
    at getStorageServiceSettings (/usr/local/lib/node_modules/azure-cli/lib/util/storage.util._js:1088:24)
    at StorageUtil.getBlobService (/usr/local/lib/node_modules/azure-cli/lib/util/storage.util._js:139:25)
    at Object.StorageUtil.getServiceClient (/usr/local/lib/node_modules/azure-cli/lib/util/storage.util._js:522:27)
    at getBlobServiceClient (/usr/local/lib/node_modules/azure-cli/lib/commands/storage/storage.blob._js:67:37)
[...]

PR #2104, opened 2015-09-15
Fix issue #2102 and #2103

yaxia:
1. Fix the issue that it reports: Cannot read property 'substr' of null when user doesn't login or no endpoint is provided.
2. Modify the build script as the npm.zip is only available in https address now.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

tools/windows/scripts/prepareRepoClone.cmd
```diff
@@ -18,8 +18,8 @@ set PATH=%PATH%;"C:\Program Files (x86)\Git\bin;"
 
 pushd %~dp0..\
 
-set NODE_DOWNLOAD_URL=http://nodejs.org/dist/v%NODE_VERSION%/node.exe
-set NPM_DOWNLOAD_URL=http://nodejs.org/dist/npm/npm-%NPM_VERSION%.zip
+set NODE_DOWNLOAD_URL=https://nodejs.org/dist/v%NODE_VERSION%/node.exe
+set NPM_DOWNLOAD_URL=https://nodejs.org/dist/npm/npm-%NPM_VERSION%.zip
```
yaxia: The reason I change it is that at the day Node.js 4.0 was released, this npm cannot be downloaded from this address and the server returned HTTP 301 (Moved Permanently), only HTTPS worked. It looks like the address works again now, but I am not sure whether it will be inaccessible again.

PR #2108, opened 2015-09-17
AzureRT S54 PR #3 - Extension Image List

huangpf:


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

test/commands/arm/vm/arm.vm-image-list-tests.js
```diff
@@ -92,6 +93,47 @@ describe('arm', function() {
...
+          done();
+        });
+      });
+
+      it('extension list-image-types ', function(done) {
+        var cmd = util.format('vm extension list-image-types %s %s --json', location, publisher).split(' ');
+        testUtils.executeCommand(suite, retry, cmd, function(result) {
+          result.exitStatus.should.equal(0);
+          var allResources = JSON.parse(result.text);
+          type = allResources[0].name;
```
yugangw-msft: do you need to check the value?

test/commands/arm/vm/arm.vm-image-list-tests.js
```diff
@@ -92,6 +93,47 @@ describe('arm', function() {
...
+          done();
+        });
+      });
+
+      it('extension list-image-versions ', function(done) {
+        var cmd = util.format('vm extension list-image-versions %s %s %s --json', location, publisher, type).split(' ');
+        testUtils.executeCommand(suite, retry, cmd, function(result) {
+          result.exitStatus.should.equal(0);
+          var allResources = JSON.parse(result.text);
+          version = allResources[0].name;
```
yugangw-msft: same here.

test/commands/arm/vm/arm.vm-image-list-tests.js
yugangw-msft: also, you will need to declare the variable "var type;", so it won't go to the global scope.

test/testlistarmvm.txt
```diff
@@ -1,6 +1 @@
-commands/arm/vm/arm.vm-create-tests.js
```
yugangw-msft: do you mean to remove all those tests?

huangpf: We reduce this to one or several tests that will be run for runners. New tests shall still be added to ```testlistarm.txt```.

test/commands/arm/vm/arm.vm-image-list-tests.js
huangpf: Actually, it's meant to be global: ```var location, publisher, offer, sku, type, version;``` at line 29.

test/commands/arm/vm/arm.vm-image-list-tests.js
huangpf: Actually, it's meant to be global: ```var location, publisher, offer, sku, type, version;``` at line 29.

huangpf: The workflow is to use the return fields of test ```i``` as for input for test ```i + 1```, so as long as they don't run in parallel, they should pass in sequence.

PR #2164, opened 2015-10-01
do not set language parameter when invoke adal-node for user code 

yugangw-msft:
Using "es-mx" is not a valid value. Once AAD server fixes the bug at its end, it will interpret as Spanish language and start to offer response in that language.


yugangw-msft:
The CI failure was caused by a known test flakiness. Issue open #2167

Issue #2167, opened 2015-10-02
Travis build failure due to 'arm group template' test cases


romangromov:
The reason is merged PR https://github.com/Azure/azure-xplat-cli/pull/2164 with failed Travis build https://travis-ci.org/Azure/azure-xplat-cli/jobs/83047750

yugangw-msft:
This is a known test flakiness. The test should not replay a public gallery template which will change and and fail test.

yugangw-msft:
Fixed through #2176

PR #2171, opened 2015-10-02
disable group template test (tracked by #2167)

yugangw-msft:


yugangw-msft:
@amarzavery please revert the change when you submit PR to fix the test issue. Please verify the test can now run w/o network connectivity which means fully mocked

PR #2174, opened 2015-10-05
[Do not merge] Enable login using MSA account such as live id

yugangw-msft:
Mark it as do not merge, because adal-node it depends on has not been published


yugangw-msft:
Closing, a new PR coming with no conflicts with dev branch.

PR #2176, opened 2015-10-07
Fix the gallery list/show/download commands 

yugangw-msft:
Please do not merge. This PR has conflict with #2174, which has higher priority. 
1. "--gallery-template" parameter are removed from the "group" or "group deployment" for the reason that the server introduced a breaking change that it no longer provides default template
2. "group template show & download" are modified to handle multiple template json files


lib/commands/arm/group/group.template._js
```diff
@@ -139,18 +138,23 @@ exports.init = function(cli) {
...
+        var fileName = path.join(directory, uris[i].name + '.json');
+        fileName = groupUtils.normalizeDownloadFileName(fileName, options.quiet, confirm, _);
+        if (fileName) {
+          log.info('Downloading ' + fileName);
+          waitForDownloadEnd(request(uris[i].uri).pipe(fs.createWriteStream(fileName)), _);
+        }
+      }
     });
 
   groupTemplate.command('validate [resource-group]')
```
vivsriaus: The validate calls the deployment Uri, so this should ideally be inside the groupDeployment commands. We made this change recently in Powershell as well (renamed Test-AzureResourceGroupTemplate to Test-AzureResourceGroupDeployment and moved it inside our deployment folder).

yugangw-msft: @vivsriaus this is about the existing logic. For the sole purpose of risk control, let us do the tweak when we start to restructure the gallery command category

PR #2178, opened 2015-10-08
Enable login using MSA account such as live id 

yugangw-msft:
Mark it as do not merge, because adal-node it depends on has not been published


yugangw-msft:
Reviewed by Amar.

PR #2190, opened 2015-10-11
Remove streamline temp folder at postinstall

yugangw-msft:
We have received at least 5 reports, that stale streamline files caused weird azure-cli errors. So we just remove the folder.


yugangw-msft:
Closing, as removing external folder won't work for OSX and LINUX because npm chooses not to run script under root, unless you specify "--unsafe-perm" when invoke "npm install", but that is might not be something we would like to tell users to do

PR #2206, opened 2015-10-19
[do not merge] AzureRT S55 - VMSS Commands

huangpf:
@yugangw-msft @amarzavery


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/util/utils.js
```diff
@@ -176,7 +176,7 @@ exports.createStorageResourceProviderClient = function (subscription) {
 };
 
 exports.createComputeResourceProviderClient = function (subscription) {
-  var factoryMethod = require('azure-arm-compute').createComputeManagementClient;
+  var factoryMethod = require('./../commands/arm/armsdk/compute').createComputeManagementClient;
```
amarzavery: Why are you putting the compute sdk code in CLI? I think the compute sdk code should go to azure-sdk-for-node and be published as a package and referenced from there.

test/commands/arm/vm/arm.vmss-create-tests.js
```diff
@@ -0,0 +1,180 @@
...
+            testUtils.executeCommand(suite, retry, cmd, function(result) {
+              result.exitStatus.should.equal(0);
+              done();
+            });
+            */
+          });
+        });
+      });
+
+      /* TODO: The following commented tests won't run. */
```
amarzavery: Can these tests be uncommented as the problem was fixed?

lib/util/utils.js
huangpf: VMSS is still in internal preview. Once it's in public preview, we will publish it as package so that we can reference it here. For now, we have to check in the code instead. Thanks

test/commands/arm/vm/arm.vmss-create-tests.js
huangpf: Should be. I will try to fix them.

amarzavery:
A general comment. Please follow the design for creating new cmdlets so as to maintain consistency.
```/* 
5610   VirtualMachineScaleSet DeallocateInstances 
5611   --resource-group-name  // This should be --resource-group and it should have a short hand switch called "-g"
5612   --vm-scale-set-name 
5613   --vm-instance-ids 
```

amarzavery:
We have never had a command named create-or-update. I would recommend either using the verb "create" or "update"

PR #2256, opened 2015-11-03
Fix for account creation and a couple renames

begoldsm:
remove 'tags' from help and syntax, which resulted in parsing issues
when generating the commands.

Account creation is now functional again and tags show up properly under
options.

Also missed a couple datalake -> datalakestore renames, those are now included.

Finally, made --force and --folder actual switches, instead of having them require a value.

### DIFF
--- a/lib/commands/arm/datalakestore/datalakeStore.utils.js
+++ b/lib/commands/arm/datalakestore/datalakeStore.utils.js
@@ -31,20 +31,27 @@ exports.showObject = function(log, object, indentationTabs) {
   var tabs = __.isNumber(indentationTabs) ? indentationTabs : 0;
   var spaces = _padSpaces(tabs);
   var recursiveCaller = function (element) { exports.showObject(log, element, tabs + 1); };
-  for (var propertyName in object) {
-    if (__.isNull(object[propertyName]) || __.isUndefined(object[propertyName])) {
-      log.data(spaces + propertyName + ':');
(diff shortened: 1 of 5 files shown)


lib/commands/arm/datalakestore/datalakeStore.utils.js
```diff
@@ -31,20 +31,27 @@ exports.showObject = function(log, object, indentationTabs) {
...
-    if (__.isNull(object[propertyName]) || __.isUndefined(object[propertyName])) {
-      log.data(spaces + propertyName + ':');
-    } else if (__.isArray(object[propertyName])) {
-      log.data(spaces + propertyName + ':');
-      __.each(object[propertyName], recursiveCaller);
-    } else if (!__.isFunction(object[propertyName])) { // Do not recurse if the object[propertyName] is a function
-      if (object[propertyName].toIsoString !== undefined) { // Special case for TimeGrain objects returned by server
-        log.data(spaces + propertyName + ': ' + object[propertyName].toIsoString());
-      } else if (__.isObject(object[propertyName])) {
+  // Test if the object to log is just a string. If so, simply log it at the current level
```
amarzavery: This seems like duplicate code. I saw similar changes in datalakeanalytics.utils.js . If that is the case then there should be a common utils.js file that has the common code. This common file can then be required in datalakestore and datalakeanalytics util files. This will reduce maintenance efforts.

This is ok for this release. Please make sure that this refactoring is done in the next release.

PR #2277, opened 2015-11-10
AzureRT PR - Package Dependency Update

huangpf:


package.json
```diff
@@ -78,8 +78,10 @@
     "kuduscript": "1.0.3",
     "mime": "~1.2.4",
     "moment": "2.6.0",
+    "ms-rest-azure": "1.2.0",
     "node-forge": "0.6.23",
     "node-uuid": "1.2.0",
+    "number-is-nan": "1.0.0",
```
yugangw-msft: confirm: do you also need to bubble this one to get installer built out at a windows machine? I thought doing it on "ms-rest-azure" could be enough.

huangpf: Yes, both are needed...

PR #2330, opened 2015-12-01
ARM: TrafficManager Profile & Endpoints changes

romangromov:
This PR includes changes related to `traffic-manager` commands in ARM.
Also, related test cases updated and recordings updated.

+@jtuliani for visibility,

+@amarzavery @yugangw-msft to merge


lib/commands/arm/network/network._js
```diff
@@ -1649,13 +1649,12 @@ exports.init = function (cli) {
...
     .option('-g, --resource-group <resource-group>', $('the name of the resource group'))
     .option('-n, --name <name>', $('the name of the profile'))
-    .option('-l, --location <location>', $('the location'))
     .option('-u, --profile-status <profile-status> ', util.format($('the profile status, valid values are' +
     '\n     [%s], default is %s'), constants.trafficManager.status, constants.trafficManager.status[0]))
     .option('-m, --traffic-routing-method <traffic-routing-method>', util.format($('the traffic routing method for the profile,' +
     '\n     valid values are [%s], default is %s'), constants.trafficManager.routingMethod, constants.trafficManager.routingMethod[0]))
     .option('-r, --relative-dns-name <relative-dns-name>', $('relative DNS name of the profile e.g. .trafficmanager.net'))
-    .option('-e, --ttl <ttl>', $('time to live in specified in seconds'))
+    .option('-l, --ttl <ttl>', $('time to live in specified in seconds'))
```
amarzavery: I think this will be a breaking change for customers who have scripts with -l switch and would be passing location over there. Removing location is understandable (as location is implied from the location of resource group). However, replacing -e with -l would break customers, if that letter was used for something else in the same command.

romangromov: @amarzavery this is requested change, comment from Jonathan:
>b.	Instead, ‘-l’ should be used as the shortcut for ‘—ttl’.  This is then consistent with ‘azure network dns record-set create’

jtuliani: Confirmed!

Issue #2268, opened 2015-11-06
Wrong graph end point for AzureChinaCloud


amanda0411:
Hi 
I see that the value of "activeDirectoryGraphResourceId" for "AzureChinaCloud" is "https://graph.windows.net/" in file envrionment.js. Actually, the value for AzureChinaCloud should be "http://graph.chinacloudapi.cn". Could you have a look? :)

yugangw-msft:
fixed it

PR #2331, opened 2015-12-01
Fix graph url endpoint for AzureChinaCloud environment

karlaugsten:
graph url should point to the China instance of graph, not public.

This causes tokens from the issuer to be invalid when presented to graph.


yugangw-msft:
fixed #2268

PR #2345, opened 2015-12-02
login: ensure we can cleanup token entries added by old version of tools

yugangw-msft:


yugangw-msft:
@amarzavery , please review once you get time.

PR #2353, opened 2015-12-03
Authorization role commands update to take scoped role definitions

shuagarw:
- Set roleDefinition to take complete object instead of partial updated fields.
 - List, show and delete role definition to take scope param
 - list and show RoleDefinition to take AtScopeAndBelow param
 - Use scoped RoleDefinitions in RoleAssignment commands


lib/commands/arm/role/role.definition._js
```diff
@@ -30,28 +30,40 @@ exports.init = function (cli) {
...
     .description($('Lists Azure RBAC roles available for assignment. Use this command to determine what actions on what resource types an Azure RBAC role allows.'))
-    .usage(
+    .usage('[scope]' +
       '\n' +
       '\n           -----    Example 1   -------' +
       '\n           azure role list' +
       '\n           Lists all RBAC role definitions.' +
       '\n')
     .option('--custom', $('If specified, display only the custom role definitions instead of all role definitions.'))
+    .option('--scope <scope>', $('Role definition scope.'))
```
amarzavery: What is scope? What are the valid values for scope. There is a big example out there but does not mention anything about scope.

shuagarw: That's fine (It's a known thing). This is for our custom roles GA, but before that happens we will update the docs and examples. As of now, existing functionality is working with the new changes and that's what we are more concerned with.

PR #2365, opened 2015-12-07
Installer: remove big VS profile temp data file 

yugangw-msft:
This shrink down the installer size by 10MB


yugangw-msft:
reviewed by Amar

PR #2395, opened 2015-12-24
Collect basic metrics when enabling vm diagnostics

colrack:
With this patch, enabling diagnostics with `azure vm enable-diag` without specifying a `wadcfg` file (with `--config-file`) will use a default generated `wadcfg` config file that will collect basic metrics.
You can have a preview in the attached zip file on what default wadcfg looks like on Windows and what it looks like on Linux; this is not something I invented: this is the default configuration that is set when you enable the diagnostics from the new portal. With this, you will be able to collect and see `basic metrics` and view charts from the new portal and you will be able to use Azure Insights. 

I have also updated the default Windows and Linux extension to the latest version available (Windows 1.5 and Linux 2.2).

Linux:
linux_basic_metrics

Windows:
windows_basic_metrics

Default wadcfg files:
[linux_windows_basic_metrics_wadcfg.zip](https://github.com/Azure/azure-xplat-cli/files/71358/linux_windows_basic_metrics_wadcfg.zip)

I did a quick test on latest images currently available (and supported by Azure) and it seems all ok (Ubuntu, Debian, CentOS, Oracle Linux, OpenSuse and Windows Server; I managed to make it work even on CoreOS but this OS is not currently supported by the diagnostics extension).

I noticed that there is some sort of confusion in the code about the `name` of the extension and the the `type` of the extension. Currently in the code there is no distinction between these. We are using the `extensionType` as the `extensionName`. This is something to be fixed up to me (I can work on it), problems can arise cause of this.
As an example: when you enable the diagnostics extension from the new portal, the extension given name is `Microsoft.Insights.VMDiagnosticsSettings` both on Windows and Linux. If you first enable the diagnostics extension from the portal, you will be not able to update because of "Multiple VMExtensions per handler not supported"; you have to remove and then reinstall the extension to update. The same is happening if you first enable from the cli: updating from the portal won't work. See attached files. This is due to the fact that when we install or update the extension from the cli, the given name for the extension is the same name of the extension type (on Windows this is `Microsoft.Azure.Diagnostics.IaaSDiagnostics` and on Linux `Microsoft.OSTCExtensions.LinuxDiagnostic`). We should let the user set extension name in both `azure vm enable-diag` and `azure vm extension set` commands and set default diagnostics extension name as the portal set: `Microsoft.Insights.VMDiagnosticsSettings`. In the code there is already some unused `--reference-name` option and I can guess what it was meant to be - this seems to be used on asm mode only.

When you enable from the portal, extension name is `Microsoft.Insights.VMDiagnosticsSettings`:
microsoft_insights_vmdiagnosticssettings

This happens when you try to update:
diag_extension_name_portal

### DIFF
--- a/lib/commands/arm/vm/vmExtensionProfile._js
+++ b/lib/commands/arm/vm/vmExtensionProfile._js
@@ -145,8 +145,12 @@ __.extend(VMExtensionProfile.prototype, {
     } else {
       if (this.params.osType === 'Windows') {
         this.cli.output.verbose($('--config-file is not specified, using default one.'));
-        config.xmlCfg = new Buffer(this._generateDefaultXmlCfg(this.params.vmID)).toString('base64');
+        config.xmlCfg = new Buffer(this._generateDefaultWindowsXmlCfg(this.params.vmID)).toString('base64');
+      } else if (this.params.osType === 'Linux') {
+        this.cli.output.verbose($('--config-file is not specified, using default one.'));
+        config.xmlCfg = new Buffer(this._generateDefaultLinuxXmlCfg(this.params.vmID)).toString('base64');
       } else {
+
         return null;
       }
     }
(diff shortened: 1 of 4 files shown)


azurecla:
Hi __@colrack__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

amarzavery:
Overall the PR looks good to me. @huangpf from the Compute team, can you please review the PR as well?

lib/commands/arm/vm/vmExtensionProfile._js
```diff
@@ -156,34 +160,918 @@ __.extend(VMExtensionProfile.prototype, {
...
         PerformanceCounters: {
           '@': {
             scheduledTransferPeriod: 'PT1M'
           },
-          PerformanceCounterConfiguration: {
-            '@': {
-              counterSpecifier: '\\\Processor(_Total)\\\% Processor Time',
-              sampleRate: 'PT1M',
-              unit: 'percent'
+          PerformanceCounterConfiguration: [
```
huangpf: Should these settings be in a separate file?

huangpf:
LGTM, if all tests passed. Thanks.

lib/commands/arm/vm/vmExtensionProfile._js
colrack: `generateDefaultXmlCfg` was already there so I just split it having as a result 2 functions. However they are a lot of lines, so maybe including in the project 2 wadcfg xml files would be better (and maybe more clear)

PR #2398, opened 2015-12-28
AzureRT PR - Boot Diagnostics Update & Add Extension Image List Commands

huangpf:


huangpf:
@amarzavery @yugangw-msft

PR #2402, opened 2016-01-04
remove '.only' usage from ad test so that all other tests can run in CI

yugangw-msft:


yugangw-msft:
@amarzavery @shuagarw FYI

PR #2422, opened 2016-01-12
AzureRT PR - Test Related Updates

huangpf:


huangpf:
@yugangw-msft @amarzavery 
Could you please help take a look?

test/testlistarmvmlive.txt
```diff
@@ -0,0 +1,13 @@
...
+commands/arm/vm/arm.vm-create-tests.js
+commands/arm/vm/arm.vm-stop-restart-tests.js
+commands/arm/vm/arm.vm.disk-attachnew-detach-test.js
+commands/arm/vm/arm.vm-create-generate-ssh-tests.js
+commands/arm/vm/arm.vm.extension-tests.js
+commands/arm/vm/arm.vm.size-tests.js
+commands/arm/vm/arm.vm-image-list-tests.js
+commands/arm/vm/arm.availset-tests.js
+commands/arm/vm/arm.vm-disk-encryption-tests.js
+# Live Only
```
yugangw-msft: @huangpf want to clarify, we discussed about *embed list*, where my understanding is to reference a file with another list of test; but that one will not handle your PR that you like to mix test mode.
It is not impossible to do, but this is making the test framework complicated w/o too much benefits.

Issue #2414, opened 2016-01-08
How do you create a VPN Gateway?


pcgeek86:
I'm trying to create a VPN Gateway, which I thought would be fairly straightforward, but I'm getting prompted for a "Private IP Address" parameter. There's no documentation for this parameter, so I have no idea how to create one.

1. Please ensure documentation is detailed.
2. Please provide an example for creating a VPN Gateway.

Cheers,
Trevor Sullivan
Microsoft MVP: PowerShell
http://trevorsullivan.net
http://twitter.com/pcgeek86

colrack:
Hi
I played with vpn connections few months ago so I can answer
Following cmds are for ARM mode

1) create a group
`azure group create vpngroup --location westeurope`

2) create a public ip for the gateway
`azure network public-ip create --resource-group vpngroup --name gwpip --location westeurope`

3) create a vnet
`azure network vnet create --resource-group vpngroup --name gwvnet --location westeurope --address-prefixes 10.0.0.0/16`

4) create a subnet for the gateway; notice as for now the name must be `GatewaySubnet` for some technical reasons
`azure network vnet subnet create --resource-group vpngroup --vnet-name gwvnet --name GatewaySubnet --address-prefix 10.0.1.0/24`

5) create the vpn gateway (this will take a little time so be patient, ~30 min)
`azure network vpn-gateway create --resource-group vpngroup --name gw --location westeurope --vnet-name gwvnet --subnet-name GatewaySubnet --private-ip-address 10.0.1.4 --public-ip-name gwpip`

The  --private-ip-address parameter is the local ip address for the gateway in the vnet; virtual machines traffic will be routed to the gw to reach VMs in another vnet or on premise network. Notice that a VM can be placed in another subnet of the same vnet (it is not necessary to place it in the `GatewaySubnet `).

Docs are poor actually but I came to a working VPN connection (obviously cmds above are not enough to establish an actual connection you have to setup at least another point to connect to).

Hope it helps
Bye

pcgeek86:
@colrack Wow, thanks so much for that. I really appreciate the help. Let's leave this issue open so that the documentation can be fixed. :) The `--private-ip-address` parameter is what confused me the most. I haven't ever seen the requirement to configure a private IP address on a VPN Gateway before. You don't should have to specify the subnet name or subnet ID, at least in theory. :)

Ciao!
Trevor Sullivan

romangromov:
[...]

PR #2427, opened 2016-01-13
ARM: 'azure network vpn-gateway create/set/list' improved

romangromov:
Minor changes, related issue fixed https://github.com/Azure/azure-xplat-cli/issues/2414

### DIFF
--- a/lib/commands/arm/network/constants._js
+++ b/lib/commands/arm/network/constants._js
@@ -87,6 +87,7 @@ module.exports = {
   },
 
   vpnGateway: {
+    subnetName: 'GatewaySubnet',
     type: ['RouteBased', 'PolicyBased', 'Dedicated'],
     connectionType: ['Vnet2Vnet', 'IPsec', 'ExpressRoute'],
     defWeight: 0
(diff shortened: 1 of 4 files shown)


colrack:
Good job, now it is possible to create a vpn gateway without providing `--subnet-name` and `--private-ip-address`. 

Suggestion: specify in description of `--private-ip-address` that it defaults to a Dynamic ip address.

Quick live test seems ok but there is a small issue. If you don't provide `--enable-bgp` you'll get:
`Error "Error: this['enableBgp'] must be of type boolean`
This is because default value is string ['false'](https://github.com/MSOpenTech/azure-xplat-cli/blob/arm-vpn-gateway-improve/lib/commands/arm/network/virtualNetworkGateway._js#L62). Solution: change it to `false`, boolean.
[Tests](https://github.com/Azure/azure-xplat-cli/blob/dev/test/commands/arm/network/arm.network.vpn-gateway-tests.js) are not catching it because `--enable-bgp` is provided and then parsed to obtain a bool value.

romangromov:
@colrack thanks a lot, `-private-ip-address` description improved, reported issue for `-enable-bgp` fixed

romangromov:
@amarzavery @yugangw-msft please merge

PR #2438, opened 2016-01-17
simplify logging by removing all hacks, patches and redundancies

yugangw-msft:
Simply the logging code and get winston do the most heavy lifting. This way the logging code gets much simpler. Also the azure-cli's silly logging (azure.detail.*) no longer needs truncation as winston's async streaming will get all drained up. To get this natural process work, I updated azure-cli's exit logic not to call process.exit unless there is error code, which is a recommended practice anyway.


yugangw-msft:
Have been reviewed by 2 people, so merging

PR #2446, opened 2016-01-19
Migrated express-route authorization on Autorest SDK

MikhailTryakhov:
Removed status options from create/set commands, from autotests


MikhailTryakhov:
Need to fix issues in unit tests

PR #2454, opened 2016-01-20
build: simplify the script of creating tarballs

yugangw-msft:
@amarzavery, please review.
1. Get rid of the linux make files, which is not useful as it is a simple wrapper around npm, nothing more.
2. Rename the "osx-setup" to "nix-setup", this deals with the confusion we always have that, why the creating linux tarball script is under osx folder?
3. Ensure the linux tarball contains the same set of file as the installer, mainly 
    a. No product code is removed
    b. streamline line files are pre-compiled


yugangw-msft:
need to confirm how the tarball to be used on linux before commit this. closing

PR #2462, opened 2016-01-21
build: revert to old ci option to see whether it gets performence better

yugangw-msft:


yugangw-msft:
the number for node (0.10, 0.12, 4.2.4) are 6m, 8m,6m. Verse 10m, 10m, and 10m this is better. So merging

PR #2463, opened 2016-01-22
AzureRT PR - update vm quick-create, use NRP autoRest-client, & other fixes

huangpf:


huangpf:
@yugangw-msft @amarzavery Could you please help take a look?

lib/commands/arm/vm/vmClient._js
```diff
@@ -201,13 +201,15 @@ __.extend(VMClient.prototype, {
     params.imageUrn = imageUrn;
     // hardwareProfile
     // For quick create requirement is to have 'Standard_D1' as default
-    params.vmSize = 'Standard_D1';
+    params.vmSize = options.vmSize ? options.vmSize : 'Standard_D1';
```
yugangw-msft: nit: you can simplify with (options.vmSize || 'Standard_D1')

lib/commands/arm/vm/vmClient._js
```diff
@@ -201,13 +201,15 @@ __.extend(VMClient.prototype, {
...
     // osProfile
     params.computerName = params.vmName;
     params.adminUsername = adminUsername;
     params.adminPassword = adminPassword;
+    params.sshPublickeyFile = options.sshPublickeyFile;
     params.osType = osType;
     // storageProfile - storage accountweb
+    params.storageAccountName = ('cli' +  hash + utils.getHash(adminUsername)).substring(0, 24);
```
yugangw-msft: i don't have the domain knowledge, but generally speaking, use hard-coded 0 and 24 to retrieve substring looks unprotected on error cases or new cases.

scripts/unit.js
```diff
@@ -38,7 +38,7 @@ function buildFileList(testFiles, testList, root) {
   var includeMark = 'include:';
   for (var i = 0; i < files.length ; i++) {
     if (files[i].indexOf(includeMark) === 0) {
-      var fileToInclude = files[i].substring(includeMark.length).replace('\r', '');
+      var fileToInclude = files[i].substring(includeMark.length).replace(/\s+$/g, '');
```
yugangw-msft: please back out the change, using "\r' " is the fix, which is consistent with the other code

yugangw-msft:
@romangromov, please take a look, this touches network code. I think it is due to expected interface change

lib/commands/arm/vm/vmClient._js
huangpf: Node.js' substring function will extract 'up to' the ```end``` position, but not including. So I think it should be fine here.

http://www.w3schools.com/jsref/jsref_substring.asp

```
Parameter	Description
start	Required. The position where to start the extraction. First character is at index 0
end	Optional. The position (up to, but not including) where to end the extraction. If omitted, it extracts the rest of the string
```

romangromov:
@yugangw-msft reviewed, this PR do not affect any `azure network` command, changes related to Networking SDK from autorest looks good to me

PR #2472, opened 2016-01-24
test: misc improvements

yugangw-msft:
1. modify provider test not to use appservice, which is going away… 
2. when run live, ensure recording mode is off


yugangw-msft:
@amarzavery , fyi

PR #2488, opened 2016-01-29
ARM network: 'vpn-connection delete' test updated

romangromov:
Minor fix for 'network vpn-connection delete' test case to run against live server


romangromov:
started issue for CI failure https://github.com/Azure/azure-xplat-cli/issues/2489

PR #2491, opened 2016-01-30
Azure CP - Fix ASM azure vm list

huangpf:
@amarzavery 
This is an issue that is found recently that breaks a very common user case of listing VMs.

The cause is on the service side that it *sometimes* does not return the DNS for a VM, but here we tried to make our client more robust.

Please consider including this into the incoming release. Thanks!


huangpf:
@sauryadas FYI.

amarzavery:
@huangpf - Please add a test to verify that it does not regress.

huangpf:
@amarzavery 
The existing test passes validated regression. In terms of negative tests, this error occurs only if the user's VM has problem, i.e. missing a required field. So, could you please help provide some instruction for us on how to mock this for both ```record``` and ```playback``` modes in the current test infrastructure? Thanks.

amarzavery:
@huangpf - How does the VM enter into a state such that the required field is missing? That would be the place you would want to fill the hole and enter a test for that. You should be having more context about that. This brings up a bigger question - How sure are we that the fix in this PR solves the issue in a foolproof way? Hence having a negative test helps.

huangpf:
It's a sporadic issue that not every subscription can reproduce it. I can record and play it back on my end. But your subscription may not have this issue.

amarzavery:
@huangpf - I see. I am merging this PR.

PR #2520, opened 2016-02-10
AzureCP PR - VM Create License Type Option

huangpf:


huangpf:
@amarzavery @yugangw-msft Could you please help review? Thanks.

package.json
```diff
@@ -12,7 +12,7 @@
     "Zavery, Amar <amzavery@microsoft.com>",
     "Wang, Yugang <yugangw@microsoft.com>"
   ],
-  "version": "0.9.15",
+  "version": "0.9.15.1",
```
huangpf: @amarzavery Is it OK to use a 4th digit for the version string here?

amarzavery: npm follows [semver](https://github.com/npm/node-semver).
I did a basic test to validate my understanding.
```cmd
notepad package.json  #edited the package.json to have 'empty object' {}
npm install semver --save
```
Create a file named test.js with following test:
```js
var semver = require('semver');
console.log('>>>>>>>>>>');
console.log(semver.valid('0.9.15'));
console.log(semver.valid('0.9.15.1'));
```

The output looks like this:
```
D:\sdk>node test.js
>>>>>>>>>>
0.9.15
null
```
This asserts that npm only allows major.minor.patch and nothing further than that.

If you cannot wait till the next release, then you can update the package.json to "0.9.16" and we will publish this to npm. We will not be generating new installers as this involves lot of work and testing. Customers can get the new version from npm.

If this is targeted for release then this PR should not be merged into the dev branch. We should create a release branch from the 0.9.15 release tag and then send a PR over there.

Why was this not communicated earlier by the compute team? I understand that there is an off band package release for powershell, however there was no communication about Xplat CLI.

huangpf: The CLI work item had not been created or being tracked, until recently mentioned. I would try to follow your instruction on the version string, and change it to ```0.9.16```. Thanks!

amarzavery:
I am closing this PR and I am creating a release branch from the previous releases tag. Please resubmit the PR in the release branch.

amarzavery:
@huangpf - Customers can download the tarball from here https://github.com/Azure/azure-xplat-cli/archive/dev.tar.gz and then do the following
- Uninstall the previously installed cli (`sudo azure-uninstall` if installed from a mac installer), (`npm uninstall azure-cli -g` if installed from npm), (unistalling the windows MSI is easy).
- `npm cache clear -g`
- `npm install -g 'path-to-downloaded-tar-ball'`
And they should be good to go.

Issue #2530, opened 2016-02-12
Login through .publishsettings breaks CLI commands


damolina-ms:
Probably same root cause as #2529 

Run: 
1. azure account download 
2.Save file (example: x.publishsettings)
3.Run: azure account import <.publishsetting file> 
4.Verify account works (example: azure account show)
5. Run: azure group create <name> <location>

**Bug**
Command fails with : _Authentication failed. The 'Authorization' header is missing._

Stack Info
    at Function.ServiceClient._normalizeError (...\npm\node_modules\azure-cli\node_modules\azure-common\lib\services\serviceclient.js:815:23)
    at ...\npm\node_modules\azure-cli\node_modules\azure-common\lib\services\filters\errorhandlingfilter.js:44:29
    at Request._callback (...\npm\node_modules\azure-cli\node_modules\azure-common\lib\http\request-pipeline.js:109:14)
    at Request.self.callback (...\npm\node_modules\azure-cli\node_modules\azure-common\node_modules\request\request.js:199:22)

yugangw-msft:
@damolina-ms, "group create" is only valid under arm mode which requires AAD token, and this can't be supplied by "importing publish settings" which only works under asm, aka rdfe.
I will take a look to see whether i can make the error better.

yugangw-msft:
fix was in

Issue #1913, opened 2015-07-13
Switching between azure config mode arm and asm


devigned:
I first login with asm via `azure account download` and `azure account import [file path]`. I run some asm commands and everything works fine. 

- Then I switch to ARM mode, I should be prompted to login if I have no Azure Active Directory credentials available.
- Upon successful `azure login`, after working in asm, I am unable to execute calls against arm as I'm greeted with the following
```bash
info:    Executing command vm image list
/error:   The server failed to authenticate the request. Verify that the certificate is valid and is associated with this subscription.
info:    Error information has been recorded to /Users/david/.azure/azure.err
error:   vm image list command failed
```
I must run `azure account clear` to remove all of my credentials and run `azure login` again to be able to use arm.

- The same thing will happen if I start in arm mode with `azure login` and shift back to asm mode.

The credential switching between these two modes is misleading and prone to failure. Users should have an intuitive experience when shifting between modes. The credentials for asm should not be mixed with arm unless they can be, and users should be warned that they will need to login with new credentials to allow access to arm. Finally, after the user logs in to the second mode, the credentials should work and not require running `azure account clear`.

huangpf:
+1

yugangw-msft:
I fixed. No "account clear"  will be needed

PR #2531, opened 2016-02-13
fix #2529, #2530, #1913 about confusions on cred shareing between asm and arm mode

### LINKED ISSUE #2529: Login through .publishsettings breaks 'azure logout' command
Repro: 

1. Run: _azure account download_
2. Save file  (example: _x.publishsettings_)
3. Run: _azure account import <.publishsetting file>_
4. Verify account works (example: _azure account show_)
5. Try to logout: _azure logout --username <username>_

**Bug **
No matter what user name you use, logout fails with 
TypeError: Cannot read property 'name' of undefined
...\azure-cli\lib\util\profile\profile.js:140:54

The problem is code assumes subscription has a defined 'user' but it seems with publishsettings is not the case.

yugangw-msft:
This improved the experiences in such aspects
1. If your current subscription is backed by certificated from a publish settings file, "azure logout" will not throw with null ref, rather tell you are not really logged in 
2. If you logged in using cert, and switch to "arm",  you will get a warning with reasonable suggestion to either run "azure login" or "azure account set"
3. If you go ahead to run arm cmdlets anyway, you will get the error with similar text
4. If you logged in under asm, and switch to arm, I verified the credential should just work to run cmdlets. 
5. If you logged in successfully, azure-cli will guarantee to select one of its subscriptions. Before, it might stick to an existing unrelated subscription, which is confusing

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -624,6 +624,10 @@ _.extend(AzureCli.prototype, {
   isAsmMode: function () {
     return utilsCore.ignoreCaseEquals(this._mode, 'asm');
   },
+  
+  isArmMode: function () {
+    return utilsCore.ignoreCaseEquals(this._mode, 'arm');
+  },
 
   checkVersion: function () {
     // Uploading VHD needs 0.6.15 on Windows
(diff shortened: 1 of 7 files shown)


yugangw-msft:
CC @amarzavery @devigned

yugangw-msft:
Please leave comments if you see any issues

Issue #2551, opened 2016-02-17
ARM: vm quick-create: succeeded without associating IP to NIC


ahmetalpbalkan:
I just created a VM from `azure vm quick-create`  but it returned successfully without associating the Public IP resource to Network Interface resource (and **therefore the output did NOT include an IP address to connect**).

My theory is some error somewhere is getting ignored at some step (as I've seen this working in the past).

I'm on v0.9.15.

```
$ azure vm quick-create ahmet-docker vm1 westus linux 'OpenLogic:CentOS:7.1:latest' alp 'myPa$$w0rd'
info:    Executing command vm quick-create
+ Looking up the VM "vm1"
info:    Using the VM Size "Standard_D1"
info:    The [OS, Data] Disk or image configuration requires storage account
+ Looking up the storage account cli1442825269193486306
info:    Could not find the storage account "cli1442825269193486306", trying to create new one
+ Creating storage account "cli1442825269193486306" in "westus"
+ Looking up the storage account cli1442825269193486306
+ Looking up the NIC "vm1-westu-1442825269-nic"
info:    An nic with given name "vm1-westu-1442825269-nic" not found, creating a new one
+ Looking up the virtual network "vm1-westu-1442825269-vnet"
info:    Preparing to create new virtual network and subnet
+ Creating a new virtual network "vm1-westu-1442825269-vnet" [address prefix: "10.0.0.0/16"] with subnet "vm1-westu-1442825269-snet" [address prefix: "10.0.1.0/24"]
+ Looking up the virtual network "vm1-westu-1442825269-vnet"
+ Looking up the subnet "vm1-westu-1442825269-snet" under the virtual network "vm1-westu-1442825269-vnet"
info:    Found public ip parameters, trying to setup PublicIP profile
+ Looking up the public ip "vm1-westu-1442825269-pip"
info:    PublicIP with given name "vm1-westu-1442825269-pip" not found, creating a new one
+ Creating public ip "vm1-westu-1442825269-pip"
+ Looking up the public ip "vm1-westu-1442825269-pip"
+ Creating NIC "vm1-westu-1442825269-nic"
+ Looking up the NIC "vm1-westu-1442825269-nic"
+ Creating VM "vm1"
+ Looking up the VM "vm1"
+ Looking up the NIC "vm1-westu-1442825269-nic"
[...]

PR #2553, opened 2016-02-18
AzureCP PR - Fix NIC-IP Association Issue #2551

huangpf:
Fix NIC-IP Association Issue #2551


huangpf:
@amarzavery @yugangw-msft Could you please help review?

PR #2585, opened 2016-02-29
AzureCP PR S58~S59: Add Plan Parameters for VM Creation

huangpf:
* Add Plan Parameters for VM Creation


huangpf:
@amarzavery @yugangw-msft Could you please help review? Thanks.

lib/commands/arm/vm/vmProfile._js
```diff
@@ -132,6 +132,11 @@ __.extend(VMProfile.prototype, {
       vmProfile.diagnosticsProfile = null;
     }
 
+    // Plan
+    if (this.params.plan !== null) {
```
yugangw-msft: simply to just "this.params.plan"; otherwise you missed the case of "undefined"

yugangw-msft:
a nit; otherwise LGTM

huangpf:
Thanks @yugangw-msft. Addressed.

PR #2596, opened 2016-03-03
Graph version update

shuagarw:
* Added public Graph client 'azure-graph'
* Upgraded graph api-version to 1.6-internal
* Fix breaking changes in graph commands


lib/util/utils.js
```diff
@@ -73,14 +73,37 @@ exports.createClient = function (factoryMethod, credentials, endpoint, options)
...
 
-exports.createAutoRestClient = function (FactoryMethod, subscription) {
+/**
+ * Create Autorest client
+ *
+ * @param {function} Method to create an instance of the client
+ * 
+ * @param {object} Subscription object
+ * 
+ * @param {object} [options] checks for isTenantBased and endpoint properties
```
amarzavery:
```
* @param {object} [options] - Optional parameters
* @param {bool} [options.isTenantBased] - A Boolean value (true) indicating that the client is tenant based
* @param {string} [options.endPoint] - The url specifying the endpoint (baseUrl) for the client
```

lib/util/utils.js
```diff
@@ -73,14 +73,37 @@ exports.createClient = function (factoryMethod, credentials, endpoint, options)
...
-  var subscriptionId = subscription.id;
-  var baseUri = exports.stringTrimEnd(subscription.resourceManagerEndpointUrl, '/');
+  var credentials = null;
+  var baseUri = null;
+  var subscriptionOrTenantId = null;
+
+  if (options && options.isTenantBased && options.isTenantBased === true && options.endPoint) {
+    subscriptionOrTenantId = subscription.tenantId;
+    credentials = subscription._createCredentials(options.endPoint);
+    baseUri = exports.stringTrimEnd(subscription.activeDirectoryGraphResourceId, '/');
```
amarzavery: baseUri = exports.stringTrimEnd(options.endpoint, '/');

test/commands/arm/ad/arm.ad-tests.js
```diff
@@ -162,11 +162,12 @@ describe('arm', function () {
         });
       });
 
-      it('should work to list and show users', function (done) {
+      it.skip('should work to list and show users', function (done) {
```
amarzavery: The ad tests should pass, correct? Why is this skipped?

lib/util/utils.js
amarzavery: We should not couple having an endpoint with isTenantBased.
if options.endPoint is provided then use it, else we default to resourceManagerEndpointUrl

lib/util/utils.js
```diff
@@ -73,14 +73,37 @@ exports.createClient = function (factoryMethod, credentials, endpoint, options)
...
+ * @returns {object} An instance of the client
+ */
+
+exports.createAutoRestClient = function (FactoryMethod, subscription, options) {
   ensureAADBackedSubscription(subscription);
-  var credentials = subscription._createCredentials();
-  var subscriptionId = subscription.id;
-  var baseUri = exports.stringTrimEnd(subscription.resourceManagerEndpointUrl, '/');
+  var credentials = null;
+  var baseUri = null;
```
amarzavery: initialize baseUri with resourceManagerEndpointUrl

shuagarw: Fixed checked in.

test/commands/arm/ad/arm.ad-tests.js
shuagarw: As we discussed this is due to a bug in autorest generated code. Will wait for fix to be in to unskip this.

Issue #2616, opened 2016-03-08
ARM: 'network dns zone import' ignores multiple records in record set of type A


romangromov:
CLI Version: **0.9.16**
OS Type: any
Installation via:  any

Mode: **ARM**

Description:
`network dns zone import` ignores multiple records in record set of type A

Steps to reproduce:
1) Used zone file:
```
test-single     1200  A  10.1.1.333
test-multiple  A  192.168.1.18
               A  10.1.1.444
```
Gives warn: `warn:    Invalid record format:                         A       10.1.1.444`

romangromov:
For visibility: @jtuliani 
Will fix

romangromov:
Fix merged, closing issue

PR #2617, opened 2016-03-08
ARM: 'network dns zone import' fixed import for record sets with multiple records of type A

romangromov:
* `network dns zone import` fixed import for record sets with multiple records of type A

Related issue fixed, please **close**  https://github.com/Azure/azure-xplat-cli/issues/2616

@amarzavery please merge


romangromov:
Need one more work here

PR #2636, opened 2016-03-15
AzureCP PR - Update Compute API Version to 2016-03-30

huangpf:
* AzureCP PR - Update Compute API Version to 2016-03-30


lib/commands/arm/vm/vmStorageProfile._js
```diff
@@ -232,12 +232,12 @@ __.extend(VMStorageProfile.prototype, {
         },
 
         _parseOSDiskParams: function(params, createOption, _) {
-          if (createOption !== 'Attach' && createOption !== 'FromImage') {
```
yugangw-msft: consider use case insensitive comparison exposed from utils.js

yugangw-msft:
a nit, the rest looks good to me

lib/commands/arm/vm/vmStorageProfile._js
huangpf: They have to match exactly. Re: https://github.com/Azure/azure-sdk-for-node/issues/1658

huangpf:
There is a regression about the enum parsing in new Node SDK. Close it for now.

huangpf:
@yugangw-msft @amarzavery 
Could you please help take a look again? I changed the client version to ```0.16.1```. Thanks.

PR #2647, opened 2016-03-17
ARM application-gateway commands and tests implemented

MikhailTryakhov:


MikhailTryakhov:
Would be recreated after migration on azure-arm-network@0.13.0

PR #2651, opened 2016-03-17
AzureCP PR : VMSS Commands

huangpf:
* +VMSS Commands


lib/util/extendedcommand.js
```diff
@@ -195,21 +195,42 @@ _.extend(ExtendedCommand.prototype, {
...
+      if (depth < 1) {
+        this.output.help('');
+        this.output.help(cat.description().cyan);
+      }
 
       if (levels === -1 || levels > 0) {
-        for (var index in cat.commands) {
-          var cmd = cat.commands[index];
-          this.output.help(' ', cmd.fullName() + ' ' + cmd.usage());
+        var startsWith = function(a, b) {
```
yugangw-msft: use utilscore.stringStartsWith

lib/util/extendedcommand.js
```diff
@@ -195,21 +195,42 @@ _.extend(ExtendedCommand.prototype, {
...
+          return a && b && a.substr(0, b.length) === b;
+        };
+        
+        // Compress command names only for 'vmss config' category, i.e.
+        // 'vmss config sku set' & 'vmss config sku delete' =>
+        // 'vmss config sku set|delete'
+        if (depth >= 1 && startsWith(cat.fullName(), 'vmss config ')) {
+          var methods = [];
```
yugangw-msft: this should be named as "commandNames"

lib/util/extendedcommand.js
```diff
@@ -195,21 +195,42 @@ _.extend(ExtendedCommand.prototype, {
...
+          for (var index2 in cat.commands) {
```
yugangw-msft: you should just names as "index"

lib/util/extendedcommand.js
```diff
@@ -195,21 +195,42 @@ _.extend(ExtendedCommand.prototype, {
...
+      depth = 0;
+    }
+
     for (var name in this.categories) {
       var cat = this.categories[name];
-
-      this.output.help('');
-
-      this.output.help(cat.description().cyan);
+      if (depth < 1) {
```
yugangw-msft: should be "depth === 0", 'depth' should never be negative

lib/util/extendedcommand.js
```diff
@@ -195,21 +195,42 @@ _.extend(ExtendedCommand.prototype, {
...
-          var cmd = cat.commands[index];
-          this.output.help(' ', cmd.fullName() + ' ' + cmd.usage());
+        var startsWith = function(a, b) {
+          return a && b && a.substr(0, b.length) === b;
+        };
+        
+        // Compress command names only for 'vmss config' category, i.e.
+        // 'vmss config sku set' & 'vmss config sku delete' =>
+        // 'vmss config sku set|delete'
+        if (depth >= 1 && startsWith(cat.fullName(), 'vmss config ')) {
```
yugangw-msft: i doubt it is necessary to introduce the depth variable, why you can't emit out the VMSS specific help information here inside the if, and then exit w/o going deeper?

lib/util/extendedcommand.js
huangpf: OK.

lib/util/extendedcommand.js
huangpf: OK.

lib/util/extendedcommand.js
huangpf: ok

lib/util/extendedcommand.js
huangpf: ok

lib/util/extendedcommand.js
huangpf: This is to group all the commands (1 & 2 levels down) at the ```vmss config``` output.

Without doing this, it would look something like this:
```
VMSS Config Help info XXX
vmss config generate
vmss config patch

VMSS Config P1 Command
vmss config p1 set|delete

VMSS Config P2 Command
vmss config p2 set|delete

...
```

lib/util/extendedcommand.js
huangpf: ok

lib/util/extendedcommand.js
huangpf: ok

lib/util/extendedcommand.js
huangpf: ok

lib/util/extendedcommand.js
huangpf: ok

lib/util/extendedcommand.js
huangpf: This is to group all the commands (1 & 2 levels down) at the ```vmss config``` output.

Without doing this, it would look something like this:
```
VMSS Config Help info XXX
vmss config generate
vmss config patch

VMSS Config P1 Command
vmss config p1 set|delete

VMSS Config P2 Command
vmss config p2 set|delete

...
```

tools/windows/Product.wxs
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
 
-  <?define ProductVersion="0.9.16" ?>
+  <?define ProductVersion="0.9.18" ?>
```
yugangw-msft: please back out the change

lib/commands/arm/vmss/vmss._js
```diff
@@ -0,0 +1,353 @@
...
+    }
+    else if (!certUtils.isX509CertPEM(sshPublickeyPemStr)) {
+      throw new Error($('Specified SSH certificate is not in PEM or SSH RSA format'));
+    }
+    
+    var sshPublickeyPemDataBase64 = certUtils.extractBase64X509CertFromPEM(sshPublickeyPemStr);
+    
+    // Set Key Variables
+    linuxConfiguration = {};
+    if (!utils.stringIsNullOrEmpty(adminPassword)) {
```
yugangw-msft: Can the whole if-stmt be simplified like:
    linuxConfiguration.disablePasswordAuthentication = utils.stringIsNullOrEmpty(adminPassword)

lib/commands/arm/vmss/vmss._js
```diff
@@ -0,0 +1,353 @@
...
+    options.osDiskCaching = 'ReadOnly';
+    options.osDiskCreateOption = 'FromImage';
+    options.osDiskName = resourceName('os-disk');
+    
+    // Setup Subscription Parameter
+    var subscription = profile.current.getSubscription(options.subscription);
+
+    // Create resource group if not exists
+    var resourceManagementClient = utils.createResourceClient(subscription);
+    var response = resourceManagementClient.resourceGroups.list({}, _);
```
yugangw-msft: why you do not use ".show"? this can save you the loop below

lib/commands/arm/vmss/vmss._js
```diff
@@ -0,0 +1,353 @@
...
+    };
+    var storageManagementClient = utils.createStorageResourceProviderClient(subscription);
+    var stoParams = {};
+    stoParams.name = generateNewStorageAccountName(name + 'sto');
+    stoParams.location = location;
+    stoParams.accountType = 'Standard_GRS';
+    storageManagementClient.storageAccounts.create(resourceGroupName, stoParams.name, stoParams, _);
+    options.storageAccountName = stoParams.name;
+    var stoNames = [ stoParams.name ];
+    for (var i = 1; i <= 4; i++) {
```
yugangw-msft: you can folder the 6 lines above to the same loop below. The only logic difference is the name, right? if yes, then it is easy to handle by using a conditional expression

lib/commands/arm/vmss/vmss._js
huangpf: ok

tools/windows/Product.wxs
huangpf: ok

lib/commands/arm/vmss/vmss._js
huangpf: ok

lib/commands/arm/vmss/vmss._js
huangpf: ok

lib/commands/arm/vmss/vmss._js
huangpf: ok

lib/commands/arm/vmss/vmss._js
```diff
@@ -0,0 +1,352 @@
...
+    }
+    else if (!certUtils.isX509CertPEM(sshPublickeyPemStr)) {
+      throw new Error($('Specified SSH certificate is not in PEM or SSH RSA format'));
+    }
+    
+    var sshPublickeyPemDataBase64 = certUtils.extractBase64X509CertFromPEM(sshPublickeyPemStr);
+    
+    // Set Key Variables
+    linuxConfiguration = {
+      disablePasswordAuthentication : !utils.stringIsNullOrEmpty(adminPassword) ? null : true
```
yugangw-msft: why this can't be "utils.stringIsNullOrEmpty(adminPassword)"? Even if 'null' is a valid value you have to use (i doubt), then this still can be "utils.stringIsNullOrEmpty(adminPassword)? true: null", because this way you don't have the negative condition which can hurt the readibility

package.json
```diff
@@ -72,6 +72,7 @@
     "event-stream": "3.1.5",
     "eyes": "0.x.x",
     "github": "0.1.6",
+    "fast-json-patch": "0.5.6",
```
yugangw-msft: Any of your new code is using this? Can't find it.

test/testlist-arm.txt
```diff
@@ -42,6 +42,8 @@ commands/arm/vm/arm.availset-tests.js
 # Extension Tests
 commands/arm/vm/arm.vm.enable-aem-tests.js
 commands/arm/vm/arm.vm-disk-encryption-tests.js
+include:./testlist-arm-vm.txt
```
yugangw-msft: you can just use the "include:./testlist-arm-vmss.txt"

yugangw-msft:
the rest look find to me.

lib/commands/arm/vmss/vmss._js
huangpf: ```null``` is to allow both password and key credentials.

package.json
huangpf: In the big ```compute._js``` file.

huangpf:
@amarzavery @haocs 
Could you please help review and merge? Thanks.

PR #2674, opened 2016-03-24
Add export resource group command

vivsriaus:
* Add group export command


lib/commands/arm/group/group._js
```diff
@@ -257,6 +259,76 @@ exports.init = function (cli) {
       });
     });
 
+  group.command('export [name] [directory]')
```
amarzavery: `[]` means the parameter is optional

lib/commands/arm/group/group._js
```diff
@@ -257,6 +259,76 @@ exports.init = function (cli) {
...
+  .description($('Captures a resource group as a template, and saves it as file on disk.'))
+  .usage('[options] <name>')
```
amarzavery: usage and the command definition should have the same number of positional parameters and should have the same convention. `<>` --> required `[]` --> optional

lib/commands/arm/group/group._js
```diff
@@ -257,6 +259,76 @@ exports.init = function (cli) {
...
+  .option('-n --name <name>', $('the resource group name'))
+  .option('-d --directory <directory>', $('the name of the destination directory. If not specified, template file will be saved in the current directory.'))
```
amarzavery: [directory] as it is optional

amarzavery: @vivsriaus - Please read this https://github.com/Azure/azure-xplat-cli/blob/dev/Documentation/Writing-Cmd.md#specifying-required-or-optional-parameters-for-a-cmdlet. It would be helpful

vivsriaus: Thanks!

lib/commands/arm/group/group._js
vivsriaus: Makes sense. Thanks

vivsriaus:
@amarzavery Thanks for the review. Addressed your comments, and as discussed, add a skipped test

PR #2675, opened 2016-03-25
Bug fix and test improvement for command "azure vm enable-aem"

amarzavery:

Add more test for command "azure vm enable-aem "
Fix storage account name case sensitive issue for command "azure vm enable-aem "


amarzavery:
Looks like the changes are important. Since @yuezh has not been responding after repeated pings. I am taking the changes in his PR and sending them here. Once this PR is merged, I shall close his PR #2663

Issue #2661, opened 2016-03-22
Validate Command Broken


sedouard:
CLI Version: **0.9.18**
OS Type: Linux Ubuntu
Installation via:  npm
Mode: ARM

Description:
`azure group template validate` fails with bad API response. Cannot read property 'providers' of undefined.

See:
https://travis-ci.org/Azure/azure-quickstart-templates/builds/117566373

Looks like the API response is bad here:
https://github.com/Azure/azure-xplat-cli/blob/v0.9.18-hotfix/lib/commands/arm/group/groupUtils._js#L312

Steps to reproduce:
1) Run `azure group create <group_name> location`
2) Then run `azure group template validate` on the elastic search template here:
https://github.com/Azure/azure-quickstart-templates/pull/1439
3) You'll get the same error as seen in the CI log.

Error stack trace:

**Please paste the content of ~/.azure/azure.err or C:\Users\username\\.azure\azure.err over here**
{ [TypeError: Cannot read property 'providers' of undefined]
  stack: [Getter/Setter],
  __frame: 
   { name: 'exports_validateTemplate__2',
     line: 107,
     file: '/app/node_modules/azure-scripty/node_modules/azure-cli/lib/commands/arm/group/groupUtils.js',
     prev: 
      { name: '__4',
        line: 161,
        file: '/app/node_modules/azure-scripty/node_modules/azure-cli/lib/commands/arm/group/group.template.js',
        prev: undefined,
        calls: 1,
        active: false,
        offset: 4,
        col: 17 },
     calls: 2,
     active: true,
     offset: 6,
     col: 33 },
  rawStack: [Getter] }
TypeError: Cannot read property 'providers' of undefined
  <<< async stack >>>
  at exports_validateTemplate__2 (/app/node_modules/azure-scripty/node_modules/azure-cli/lib/commands/arm/group/groupUtils.js:113:34)
  at __4 (/app/node_modules/azure-scripty/node_modules/azure-cli/lib/commands/arm/group/group.template.js:165:18)
  <<< raw stack >>>
    at getTemplateProviders (/app/node_modules/azure-scripty/node_modules/azure-cli/lib/commands/arm/group/groupUtils.js:312:46)
    at ___ (/app/node_modules/azure-scripty/node_modules/azure-cli/lib/commands/arm/group/groupUtils.js:118:38)
    at ___ (/app/node_modules/streamline/lib/callbacks/runtime.js:111:13)
[...]

PR #2677, opened 2016-03-25
Fix template validate when response has error

vivsriaus:
* Throw error when template validate fails

### DIFF
--- a/lib/commands/arm/group/groupUtils._js
+++ b/lib/commands/arm/group/groupUtils._js
@@ -331,7 +331,9 @@ function getDeploymentLogs(subscription, correlationId) {
 function getTemplateProviders(validationResponse) {
   if (validationResponse.properties && validationResponse.properties.providers) {
     return __.map(validationResponse.properties.providers, function (provider) {
-      return provider.namespace.toLowerCase();
+      if (provider.namespace) {
+        return provider.namespace.toLowerCase();
+      }
     });
   }
 }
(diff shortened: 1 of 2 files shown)


lib/commands/arm/group/groupUtils._js
```diff
@@ -320,9 +329,11 @@ function getDeploymentLogs(subscription, correlationId) {
 }
 
 function getTemplateProviders(validationResponse) {
-  return __.map(validationResponse.properties.providers, function (provider) {
-    return provider.namespace.toLowerCase();
-  });
+  if (validationResponse.properties && validationResponse.properties.providers) {
+    return __.map(validationResponse.properties.providers, function (provider) {
+      return provider.namespace.toLowerCase();
```
amarzavery: AS per the spec, https://github.com/Azure/azure-rest-api-specs/blob/master/arm-resources/resources/2016-02-01/swagger/resources.json#L2316  namespace is not a required property of the provider. If the namespace is null or undefined then we will get an error "Cannot execute .toLowerCase() of undefined". Please check if it exists then call the toLowerCase()

amarzavery:
Fixes https://github.com/Azure/azure-xplat-cli/issues/2661

lib/commands/arm/group/groupUtils._js
vivsriaus: Great catch! Thanks Amar

PR #2695, opened 2016-03-29
Last pull request before build

gucalder:
* 
* 
*


lib/commands/arm/insights/insights.autoscale.notifications.js
```diff
@@ -1,4 +1,4 @@
-/**
+/**77
```
amarzavery: 77 ??

gucalder: Hmm... very odd.

amarzavery:
@gucalder 

Can you please make sure that you are using [moment version greater than 2.8.0](https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/insights/package.json#L37). [**`.toIsoString()` is now deprecated. It is replaced with `.toISOString()`**](https://github.com/moment/moment/pull/1759). This conflicts with the version of moment used by AutoRest generated clients and other commands.

This file in the node sdk https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/insights/lib/insightsManagementClient.js has lot of references to it. 

Please make changes to the **`.toIsoString()`** referenced in the following files as well:
```
- azure-xplat-cli\lib\commands\arm\insights\insights.autoscale.rule.js
- azure-xplat-cli\lib\commands\arm\insights\insights.diagnostic.set._js
- azure-xplat-cli\lib\commands\arm\insights\insights.metrics._js
- azure-xplat-cli\lib\commands\arm\insights\insights.utils.js
```

gucalder:
I cannot fix the insightsManagementClient.js, it is generated code and that would require many changes in the Hydra, and need republishing of the SDK for node. But, I fixed the references in the CLI code.

amarzavery:
As I said in the email, you need to move to autorest. The command execution will still complain.

test/commands/arm/insights/arm.insights.diagnostic.set-tests.js
```diff
@@ -60,7 +60,7 @@ describe('arm', function () {
       });
 
       describe('set', function() {
-        it('should work enable all', function (done) {
+        it.skip('should work enable all', function (done) {
```
amarzavery: What is the point of skipping those tests? If the users will use the commands it will fail on them.

gucalder: These are Vinicius' tests. We figure out that the call to "toIsoString()" used in his commands, does not come from "moment". After I changed it back to lowercase, the commands and the tests started working again.

gucalder: There will still be one deprecation message in the tests. It comes from insightsManagementClient.js. I need to fix that in a separate change (with its own work item) since it originates from the Hydra specs.

gucalder:
Now the SDK for node has been updated to use moment@2.12.0 (or the latest), the command execution should not complain.
We are definitely going to do the change to autorest. It will take some time before we are done with that.

PR #2715, opened 2016-04-06
Minor change to one of the commands to infor the user of unsupported features.

gucalder:
* 
* 
*


gucalder:
It was targeted to the wrong branch.

PR #2722, opened 2016-04-07
Add policy definition and assignment commands

vivsriaus:
*  Add policy definition commands 
* Add policy assignment commands


vivsriaus:
@amarzavery Still working on tests - I'll push the test files once I have the policy assignments commands done as well.

lib/commands/arm/policy/policy.definition._js
```diff
@@ -0,0 +1,180 @@
...
+
+  var policy = cli.category('policy');
+  var definition = policy.category('definition')
+      .description($('Commands to manage your policy definitions.'));
+
+  definition.command('create [name] [policy]')
+    .description($('Creates a new policy definition'))
+    .usage('[options] <name> <policy>')
+    .option('-n --name <name>', $('the policy definition name'))
+    .fileRelatedOption('-p --policy <policy>', $('the rule for policy definition. This should be a path to a file name containing the rule.'))
```
amarzavery: How many rules can one define inside the file? Can it just be plain text or in json format?
Can you provide a link to some msdn documentation that shows how to define policy rules. This will be very helpful for our customers.

lib/commands/arm/policy/policy.definition._js
```diff
@@ -0,0 +1,180 @@
...
+    .option('--policy-string <policyString>', $('the rule for policy definition as a string.'))
```
amarzavery: in JSON format?

vivsriaus: Yep, addressed this in my last commit

lib/commands/arm/policy/policy.definition._js
vivsriaus: I've provided a link. Writing a policy rule is not straight fwd, so the acom link here would be much more helpful. Thanks for the suggestion!

PR #2728, opened 2016-04-09
AzureCP 4/15 Release

huangpf:


lib/commands/arm/availabilityset/availset._js
```diff
@@ -18,6 +18,8 @@ exports.init = function (cli) {
     .option('-l, --location <location>', $('the location'))
     .option('-t, --tags <tags>', $('the semicolon separated list of tags'))
     .option('-s, --subscription <subscription>', $('the subscription identifier'))
+    .option('-u, --platform-update-domain-count <platformUpdateDomainCount>', $('Set the platform-update-domain-count value.'))
```
amarzavery: @huangpf -u is reserved for username. Please use something else.

PR #2733, opened 2016-04-11
Batch Linux support

jasper-schneider:
* Added support for creating and displaying Batch Linux pools
* Updated the batch node-user create and set commands to support SSH keys for Linux users
* Added the batch pool list-node-agent-skus command
* Added the batch node get-remote-login-settings command


lib/commands/batch/batch.node._js
```diff
@@ -135,6 +137,13 @@ exports.init = function(cli) {
     .option('-q, --quiet', $('overwrite the destination Remote Desktop Protocol file without confirmation'))
     .appendBatchAccountOption()
     .execute(getRemoteDesktop);
+      
+  node.command('get-remote-login-settings [pool-id] [node-id]')
```
amarzavery: @jasper-schneider - This design is incorrect. Please review the [command design guidelines](https://github.com/Azure/azure-xplat-cli/blob/dev/Documentation/Writing-Cmd.md#command-design-guidelines).Just like powershell, the CLI has some rules that need to be followed.
We need to provide a consistent user experience across all the commands. 

The following format is plausible:
`azure node remote-login-settings show`.

jasper-schneider: Sure, pushed a fix.

PR #2753, opened 2016-04-13
Fix batch account command bug and add application related commands

xingwu1:
This line should add to changelog.md under Batch section (I already added)

Added the batch application package related commands


amarzavery:
@xingwu1 - Please update the description. That will go in the changelog.

xingwu1:
This line should add to changelog.md under Batch section (I already added)

Added the batch application package related commands

PR #2761, opened 2016-04-14
AzureCP PR

huangpf:


lib/commands/arm/vmss/vmss._js
```diff
@@ -206,7 +206,7 @@ exports.init = function (cli) {
     options.skuName = options.vmSize || 'Standard_D1';
     options.skuTier = 'Standard';
     options.upgradePolicyMode = 'Manual';
-    options.computerNamePrefix = name;
+    options.computerNamePrefix = name.substring(9);
```
amarzavery: What is the logic behind choosing the string from the 10th position onwards. Can the name be less than 10 characters ?

huangpf: It has to be less than 10 characters...
```
Error: Windows computer name prefix cannot be more than 9 characters long, be entirely numeric, or contain the following characters: ` ~ ! @ # $ % ^ & * ( ) = + _ [ ] { } \ | ; : . ' " , < > / ?.
```

amarzavery: Thanks for the info.

PR #2766, opened 2016-04-15
AzureCP : Fix ACS/VMSS Command Design & ComputerNamePrefix Issues

huangpf:

Fix VMSS Computer Name Issue
Fix AvailSet UD/FD Input Issue
Azure Container Service Commands


ChangeLog.md
```diff
@@ -1,4 +1,12 @@
 ##2016-04-19 Version 0.9.20
+* Fix computer name prefix issue in vmss quick-create command
+* Fix FD/UD parameter issue in availset create command
+* Added Azure Container Service set of CLI commands.
+  container config generate/patch/set/delete
```
amarzavery: Formatting. Sub bullet points
* Added Azure Container Service set of CLI commands. 
  * container config generate/patch/set/delete
  * container create-or-update [options] <resource-group-name> <container-service-name> <parameters> 
  * container delete [options] <resource-group-name> <container-service-name> 
  * container get [options] <resource-group-name> <container-service-name> 
  * container list [options] <resource-group-name>

lib/commands/arm/compute/compute._js
```diff
@@ -914,25 +914,25 @@ exports.init = function (cli) {
...
-  containerServiceCreateOrUpdate.command('create-or-update [resource-group-name] [container-service-name] [parameters]')
+  containerServiceCreateOrUpdate.command('create [resource-group] [name] [parameters]')
   .description($('The operation to create or update a container service.'))
-  .usage('[options] <resource-group-name> <container-service-name> <parameters>')
-  .option('-g, --resource-group-name <resource-group-name>', $('resource-group-name'))
-  .option('--container-service-name <container-service-name>', $('container-service-name'))
+  .usage('[options] <resource-group> <name> <parameters>')
+  .option('-g, --resource-group <resource-group>', $('resource-group'))
+  .option('-n, --name <name>', $('name'))
   .option('-p, --parameters <parameters>', $('parameters'))
```
amarzavery: A string of parameters in **JSON** format

lib/commands/arm/compute/compute._js
```diff
@@ -946,7 +946,7 @@ exports.init = function (cli) {
     cli.output.verbose('parametersObj = ' + JSON.stringify(parametersObj));
```
amarzavery: Please wrap JSON.parse and JSON.stringify statements in a try catch. They throw nasty errors that are hard to figure out. Catching them and sending meaningful errors in the callback will provide users better information

lib/commands/arm/compute/compute._js
```diff
@@ -914,25 +914,25 @@ exports.init = function (cli) {
...
+  containerServiceCreateOrUpdate.command('create [resource-group] [name] [parameters]')
   .description($('The operation to create or update a container service.'))
-  .usage('[options] <resource-group-name> <container-service-name> <parameters>')
-  .option('-g, --resource-group-name <resource-group-name>', $('resource-group-name'))
-  .option('--container-service-name <container-service-name>', $('container-service-name'))
+  .usage('[options] <resource-group> <name> <parameters>')
+  .option('-g, --resource-group <resource-group>', $('resource-group'))
+  .option('-n, --name <name>', $('name'))
   .option('-p, --parameters <parameters>', $('parameters'))
   .option('--parameter-file <parameter-file>', $('the input parameter file'))
```
amarzavery: Do you expect a json string or a json object in the file. Please mention it in the description?

lib/commands/arm/compute/compute._js
```diff
@@ -2382,64 +2382,64 @@ exports.init = function (cli) {
...
-  containerServiceDelete.command('delete [resource-group-name] [container-service-name]')
+  containerServiceDelete.command('delete [resource-group] [name]')
   .description($('The operation to delete a container service.'))
-  .usage('[options] <resource-group-name> <container-service-name>')
-  .option('-g, --resource-group-name <resource-group-name>', $('resource-group-name'))
-  .option('--container-service-name <container-service-name>', $('container-service-name'))
+  .usage('[options] <resource-group> <name>')
+  .option('-g, --resource-group <resource-group>', $('resource-group'))
+  .option('-n, --name <name>', $('name'))
   .option('--parameter-file <parameter-file>', $('the input parameter file'))
```
amarzavery: @huangpf - I see options.parameterFile being checked and passed to sdk method in the create command. However, I don't see it being checked or passed to the sdk method in delete, show and list command.
 Is this parameter switch used by the command? Am I missing something here?

huangpf: This is a bug that needs to be addressed later, as it needs more time to investigate and it was not in the last-moment fix plan.

lib/commands/arm/compute/compute._js
huangpf: I would plan to fix this, but the urgent fix as we discussed shall be for interface only, i.e. CLI method and option names.

lib/commands/arm/compute/compute._js
amarzavery: Just a suggestion, if it is not going to be used in delete, show, list then that optional parameter switch can be deleted. As can be seen from the sdk methods over [here](https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/computeManagement2/lib/operations/containerServiceOperations.js#L436). However, this can be fixed later. For now if the user provides any value to the switches it will be a no-op.

huangpf: OK. Will try to address this.

amarzavery:
@huangpf - Thanks for addressing the review feedback. Highly appreciated 👍 .

PR #2773, opened 2016-04-19
Batch job and schedule cmds

jasper-schneider:
* Added the following Batch job schedule commands
    *  azure batch job-schedule enable
    *  azure batch job-schedule disable
    *  azure batch job-schedule terminate  
* Added the following Batch job commands
    *  azure batch job enable
    *  azure batch job disable
    *  azure batch job terminate
    *  azure batch job prep-and-release-task-status list


lib/commands/batch/batch.job._js
```diff
@@ -80,6 +82,36 @@ exports.init = function(cli) {
...
+    .option('-i, --id <jobId>', $('the id of the job'))
+    .option('--disable-option <disableOption>', $('specifies what to do with active tasks during a disable job operation, available options include "requeue", "terminate", and "wait"'))
+    .appendCommonHeaderFilterOption(true, true)
+    .appendBatchAccountOption()
+    .execute(disableJob);
+
+  job.command('terminate [jobId]')
+    .description($('Terminates the specified job, marking it as completed'))
+    .option('-i, --id <jobId>', $('the id of the job'))
+    .option('--terminate-reason [terminateReason]', $('the text you want to appear as the job\'s terminate reason'))
```
amarzavery: --reason. //To keep it simple. The help text should make it clear

lib/commands/batch/batch.job._js
```diff
@@ -80,6 +82,36 @@ exports.init = function(cli) {
...
+
+  job.command('disable [jobId]')
+    .description($('Disables the specified job.  Disabled jobs do not run new tasks, but may be re-enabled later.'))
+    .option('-i, --id <jobId>', $('the id of the job'))
+    .option('--disable-option <disableOption>', $('specifies what to do with active tasks during a disable job operation, available options include "requeue", "terminate", and "wait"'))
+    .appendCommonHeaderFilterOption(true, true)
+    .appendBatchAccountOption()
+    .execute(disableJob);
+
+  job.command('terminate [jobId]')
```
amarzavery: is the jobId required? If so then it should be in angle brackets and not in square brackets

lib/commands/batch/batch.job._js
```diff
@@ -39,6 +39,8 @@ exports.init = function(cli) {
 
   var job = batch.category('job').description($('Commands to manage your Batch jobs'));
 
+  var jobPrepAndReleaseStatus = job.category('prep-and-release-task-status').description($('Commands to manage the status of your job preparation and release tasks'));
```
amarzavery: We need a better naming. This long naming is inconvenient and against the  design guidelines.

lib/commands/batch/batch.jobSchedule._js
```diff
@@ -81,6 +81,27 @@ exports.init = function(cli) {
     .appendBatchAccountOption()
     .execute(updateJobSchedule);
 
+  jobSchedule.command('enable [jobScheduleId]')
```
amarzavery: job and schedule should be separate words. This makes schedule a sub category inside. It will be logical for customers to find schedule inside job.

amarzavery: is Id required?, If so then it should be in angle brackets

lib/commands/batch/batch.jobSchedule._js
```diff
@@ -427,4 +448,160 @@ exports.init = function(cli) {
...
+    }
+    if (options.ifUnmodifiedSince) {
+      batchOptions.jobScheduleDisableOptions.ifUnmodifiedSince = options.ifUnmodifiedSince;
+    }
+
+    startProgress(tips);
+
+    try {
+      client.jobSchedule.disable(jobScheduleId, batchOptions, _);
+    } catch (err) {
```
amarzavery: This is async. I am not sure if wrapping the try catch around it like this will work?
@yugangw-msft - Will this work?

lib/commands/batch/batch.job._js
jasper-schneider: This is how the REST API is described: https://msdn.microsoft.com/en-us/library/azure/mt282170.aspx
I'm not really sure how to condense it while still having an accurate description of the API though.  I'm open to suggestions.

lib/commands/batch/batch.jobSchedule._js
jasper-schneider: Job schedules aren't a subcategory of jobs though.  In the Batch Service, a Job and a Job Schedule are both considered top level entities.  A Batch customer wouldn't expect job schedules to be underneath jobs, so that's why Xing and I opted for the job-schedule naming.

lib/commands/batch/batch.job._js
jasper-schneider: Will fix

lib/commands/batch/batch.jobSchedule._js
jasper-schneider: I'll fix the Id angle brackets though.

lib/commands/batch/batch.job._js
amarzavery: How about `azure batch job prepare-task (list/show/etc.)` OR `azure batch job prep-task list`

lib/commands/batch/batch.jobSchedule._js
amarzavery: I see.

lib/commands/batch/batch.jobSchedule._js
jasper-schneider: This works.  Here's the output when I supply a job id that doesn't exist:

D:\azure-xplat-cli-0 [batchJobAndScheduleCmds]> node bin\azure batch job enable abcd
info:    Executing command batch job enable
+ Enabling job abcd
error:   Job abcd does not exist
info:    Error information has been recorded to C:\Users\jaschnei\.azure\azure.err
error:   batch job enable command failed

lib/commands/batch/batch.job._js
jasper-schneider: Will update

lib/commands/batch/batch.job._js
jasper-schneider: I think we at least need both "prep" and "release" in the name.  For better or worse, the single API covers both job prep task results and job release task results, and omitting one from the name seems inaccurate.  

The reason why I chose to include "status" was because I thought the name would be misleading without it.  The API isn't returning the job prep task and job release task definitions.  It's only returning the state and execution results of each.  Perhaps we'll just have to depend on the help text though for that distinction.  

How does "prep-release-status" sound to you?

lib/commands/batch/batch.job._js
```diff
@@ -39,6 +39,8 @@ exports.init = function(cli) {
 
   var job = batch.category('job').description($('Commands to manage your Batch jobs'));
 
+  var jobPrepAndReleaseStatus = job.category('prep-release-status').description($('Commands to manage the status of your job preparation and release tasks'));
```
amarzavery: This looks good 👍

lib/commands/batch/batch.jobSchedule._js
yugangw-msft: That is fine. Because it is in a streamline file, all things appear in sync. The callback has been stripped out of your hand, so try-catch is the only option you have.

lib/commands/batch/batch.job._js
amarzavery: A small nit: can you move .description on another line. I am sorry, it caught my attention this time.

jasper-schneider: Sure, no worries :)

lib/commands/batch/batch.job._js
```diff
@@ -80,6 +82,32 @@ exports.init = function(cli) {
...
+
+  job.command('disable <jobId>')
+    .description($('Disables the specified job.  Disabled jobs do not run new tasks, but may be re-enabled later.'))
+    .option('--disable-option <disableOption>', $('specifies what to do with active tasks during a disable job operation, available options include "requeue", "terminate", and "wait"'))
+    .appendCommonHeaderFilterOption(true, true)
+    .appendBatchAccountOption()
+    .execute(disableJob);
+
+  job.command('terminate <jobId>')
+    .description($('Terminates the specified job, marking it as completed'))
```
amarzavery: We have start/stop as a set of verbs.
`azure site stop` (asm mode)
`azure group deployment stop`  (arm)
`azure vm stop` (arm)
`azure site job stop` (asm)
Can we have `azure batch job stop`

jasper-schneider: I went with "terminate" because that's how our REST API is labeled, and it's actually the token that appears in the URL
https://msdn.microsoft.com/en-us/library/azure/mt282181.aspx

That said, if you think "stop" would be better for a consistent CLI experience, I'm fine with changing it.  Just double checking though, would it be ok to have a "stop" without a "start"?  There's no "start job" counterpart in our API.

jasper-schneider: Converted to "stop"

amarzavery:
LGTM. Will merge once the CI passes.

Issue #2615, opened 2016-03-08
Failed ARM commands on windows return exit code 0


robbyt:
CLI Version: **0.9.16**
OS Type: Windows 7 
Installation via:  win installer

Mode: **ARM**

Description:
When a command fails, it should return a non-zero exit code. This is broken in 0.9.16:

```
C:\Users\rob>azure group list
info:    Executing command group list
+ Listing resource groups
error:   We don't have a valid access token. Please run "azure login" again.
info:    Error information has been recorded to C:\Users\rob\.azure\azure.err
error:   group list command failed


C:\Users\rob>echo Exit Code is %errorlevel%
Exit Code is 0
```

Steps to reproduce:
1) set azure cli to arm mode
2) run a command that should fail (example above)
2) text exit code

robbyt:
A bit more information on this.

When a command line program runs successfully, it should return an exit code of 0, if there are errors, it should return another exit code to communicate information about the error (the most common is 1).

The behavior of the azure cli is inconsistent on Windows. On non-Windows platforms (Linux/OSX), the azure cli returns exit code 1, when there are errors. However, on Windows the azure cli always returns exit code 0 regardless if there's an error.

The reason why people use cli tools is for scripting & automation purposes. Since the azure cli on Windows does not return a valid exit code, it makes scripting very difficult.

amarzavery:
Thanks for reporting this issue. This has been fixed now.
```
D:\sdk\azure-xplat-cli>azure group list
info:    Executing command group list
+ Listing resource groups
error:   We don't have a valid access token. Please run "azure login" again.
info:    Error information has been recorded to C:\Users\amzavery\.azure\azure.err
error:   group list command failed

D:\sdk\azure-xplat-cli>echo Exit Code is %errorlevel%
Exit Code is 1
```

PR #2784, opened 2016-04-21
fixes #2155 and #2785

### LINKED ISSUE #2155: ARM/ASM: Inconsistent exit codes upon errors
Haven't dug so deeply however I can tell certain errors do not exit with a non-zero exit code upon failure:

```
$ azure vm foo; echo ExitCode=$?
error:   'foo' is not an azure command. See 'azure help'.
ExitCode=0
```

vs

```
$ azure account set; echo ExitCode=$?

  error: missing required argument `subscriptionNameOrId'

ExitCode=1
```

This is not ideal. Imagine someone having a script like:

```bash
doThis && \
    doThat && \
    azure account seth <subs_id> && \
    dropNuclearBomb
```

the error will not be effective and script will keep going instead of exiting.

Also an example from `git`:

```
$ git foo; echo ExitCode=$?
git: 'foo' is not a git command. See 'git --help'.

Did you mean this?
	log
ExitCode=1
```

### LINKED ISSUE #2785: `azure group delete` command doesn't support asynchronous operation.
CLI Version: **0.9.17**
OS Type: Linux (Ubuntu)
Installation via:  npm

Mode: **ARM**

Description:
`azure group delete` command doesn't support asynchronous operation. My expectation is that the deleting command returns right after it's executed. Then I can do other operations instead of waiting the command finishes. Maybe after several minutes, I can run `azure group list or show` to poll the state of the group to make sure it's deleted.

Steps to reproduce:
1) Run `azure group delete  ...`

amarzavery:
* fixes #2155 and may be #2615
* fixes #2785


ahmetalpbalkan:
@amarzavery I'm often confused when you fix multiple independent issues in a single commit. going further, I advise keeping one PR to fix one issue (or multiple duplicate/closely-related issues).

PR #2795, opened 2016-04-22
AzureCP PR #2 - Split compute._js into files

huangpf:
* 	Split compute._js into files


huangpf:
@amarzavery As you can see, splitting the large file into multiple doesn't necessarily solve the problem, because sub files can be also very large. Also, if I do need things to be shown in one file, I should be able to keep it so. But GitHub won't honor that...

PR #2797, opened 2016-04-22
Regenerated help usage for webapp commands

haocs:
*  revert 'enclosed required args in angular brackets for webapp command help'
*  added help usage for webapp commands
*  regenerated help-plugins.


amarzavery:
@haocs - plugins.arms.json ?

haocs:
@amarzavery no change applied to plugin.arms.json after regenerate. I added usage() to each webapp command so that 'resource-group' is showing as required(enclosed in angular brackets) in help usage.

PR #2815, opened 2016-04-26
Update for strorage account commands

yaxia:
* Update dependency of 'azure-arm-storage' to 0.13.0-preview
* Replace '--type' by '--sku-name' in 'azure storage account create' and 'azure storage account set'
* Add '--access-tier' to 'azure storage account create' and 'azure storage account set'
* Add '--enable-encryption-service' to 'azure storage account create' and 'azure storage account set'
* Add '--kind' to 'azure storage account create'
* Add '--disable-encryption-service' to 'azure storage account set'


yaxia:
The update is to support encryption@Rest and Cool Storage which are very important features. Please help get it in this release. Thanks.

PR #2823, opened 2016-04-27
AzureCP PR : Premium VM & ACS/Migration Command Update

huangpf:
* Premium VM
* Update ARM ACS Commands
* Update ASM Migration Commands


lib/commands/arm/vm/vmClient._js
```diff
@@ -193,6 +193,45 @@ __.extend(VMClient.prototype, {
     }
   },
 
+  getDefaultVmSize: function(location, _) {
+    var subscription = profile.current.getSubscription(this.subscription);
+    var serviceClients = this._getServiceClients(subscription);
+    var virtualMachine = new VirtualMachine(this.cli, serviceClients);
+    var sizeResult = virtualMachine.getVMSizesByLocationName(location, _);
+    var i = null;
+    for (i in sizeResult) {
```
amarzavery: As per [this document for javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in)
> Because the order of iteration is implementation-dependent, iterating over an array may not visit elements in a consistent order. Therefore it is better to use a for loop with a numeric index (or Array.prototype.forEach() or the for...of loop) when iterating over arrays where the order of access is important.

Hence, I would suggest using forEach() or the classic for loop over the for .. in loop.


Another point, is there a specific reason to not check for all the names in **one condition and in the same loop**. 
```
if (sizeResult && Array.isArray(sizeResult) {
  sizeResult.forEach(function (item) {
    if (item.name === 'Standard_DS1' || item.name === 'Standard_D1' || item.name === 'Standard_D1_v2') { 
      return item.name;
    }
  }
}
```

lib/commands/asm/compute/deployment._js
```diff
@@ -99,7 +99,7 @@ var $ = utils.getLocaleString;exports.init = function (cli) {
   --service-name
   --name
 */
-  var deploymentCommitMigration = cli.category('compute').description('Commands to invoke service management operations.').category('deployment')
+  var deploymentCommitMigration = cli.category('service').description('Commands to invoke service management operations.').category('deployment')
   .description($('Commands to manage your deployment.  '));
   deploymentCommitMigration.command('commit-migration [service-name] [name]')
```
amarzavery: @huangpf, @sauryadas - the commands have not yet been changed as discussed.
`azure service deployment migration commit/prepare/abort`. Please update this as discussed.

lib/commands/arm/vm/vmClient._js
huangpf: OK. Thanks.

lib/commands/asm/compute/deployment._js
huangpf: Hi @amarzavery, there will be more deployment related commands, so we think it's better to keep it as verb-noun pattern. Otherwise, we will be having more and more sub categories...

Also, sub-category seems to indicate a sub-resource, i.e. ```vm extension``` means ```extension``` that belongs to ```vm```, so I think actually ```commit/prepare/abort-migration``` are more accurate.

@sauryadas @mahthi Thoughts?

mahthi:
Ideally, I want these commands to be changed as well. However, I need these to ship ASAP since we are lagging behind already and I want to catch this train. With that in mind, I'm okay with shipping this experience for the classic stack.

Issue #2810, opened 2016-04-25
group export: should export indented/readable JSON


ahmetalpbalkan:
CLI Version: **0.9.20**
OS Type: Mac

Mode: **ARM**

I just ran `azure group export <rg> <dir>` and this created me a 8 KB json file **with a single line**.

This command should create JSON files that are well indented and readable as ARM templates are meant to read by humans as well. I don't think there is any argument here that favors disk-space over readability.

amarzavery:
@ravbhatnagar @vivsriaus - Can you take a look at this issue?

I think `JSON.stringify(object)` before writing it to the file does that.

If you use the 3rd parameter to JSON.stringify() that should pretty print it.
`JSON.stringify(object, null, 2)`// 2 spaces
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

ravbhatnagar:
@vivsriaus - lets fix this. Agreed the output should be human readable.

ahmetalpbalkan:
thanks!

PR #2828, opened 2016-04-27
Prettify json output before writing to file for RG export and deployment save commands

vivsriaus:
* Prettify json output before writing to file for RG export and deployment save commands


amarzavery:
Fixes https://github.com/Azure/azure-xplat-cli/issues/2810

Issue #1956, opened 2015-07-24
ARM: `group deployment create` bad JSON file causes a vague error message


ahmetalpbalkan:
I have an `azuredeploy.parameters.json` file which has bad JSON syntax in it, but I don't know yet.

Then I'm running and getting a vague error message like:

```
$ azure group deployment create -g ahmetb-swarm-0 -n swarm-0 -f azuredeploy.json -e azuredeploy.parameters.json
info:    Executing command group deployment create
error:   Unexpected token a
info:    Error information has been recorded to /Users/alp/.azure/azure.err
error:   group deployment create command failed
```

I think we should simply parse those two files once and see if the JSON parsing fails. 

I opened the issue originally without knowing my JSON file was malformed and this error message led to a lot of confusion and I thought CLI was buggy.

ahmetalpbalkan:
Happens in full parameter names as well...

```
azure group deployment create \
  --resource-group ahmetb-swarm-0 \
  --template-file azuredeploy.json \
  --parameters-file azuredeploy.parameters.json
```

devigned:
Looks like it's choking on the parameters.json when it's parsing. At least that's my guess based only on the stack trace.

ahmetalpbalkan:
@devigned you're right looks like JSON syntax was invalid. We should probably come up with a better error message here. Editing the issue.

devigned:
Yes, absolutely. The error needs to be helpful.

ravbhatnagar:
Fixed

PR #2842, opened 2016-05-02
Add jsonLint.js to display the line number in exception when parsing json fails

vivsriaus:
* Add jsonLint.js to display the line number in exception when parsing json fails
* Fixes issue [1956](https://github.com/Azure/azure-xplat-cli/issues/1956)


lib/util/jsonLint.js
```diff
@@ -0,0 +1,440 @@
...
+          lstack = [], // location stack
+          table = this.table,
+          yytext = '',
+          yylineno = 0,
+          yyleng = 0,
+          recovering = 0,
+          TERROR = 2,
+          EOF = 1;
+      
+      //this.reductionCount = this.shiftCount = 0;
```
gandhiniraj: Remove?

lib/util/jsonLint.js
```diff
@@ -0,0 +1,440 @@
...
+          break;
+        case 2: this.$ = Number(yytext);
+          break;
+        case 3: this.$ = null;
+          break;
+        case 4: this.$ = true;
+          break;
+        case 5: this.$ = false;
+          break;
+        case 6: return this.$ = $$[$0 - 1];
```
gandhiniraj: Good to add comment on cases below

lib/util/jsonLint.js
```diff
@@ -0,0 +1,440 @@
+﻿/* jshint ignore:start */
+/* Jison generated parser */
+var jsonlint = (function () {
+  var parser = {
+    trace: function trace() { },
+    yy: {},
+    symbols_: { "error": 2, "JSONString": 3, "STRING": 4, "JSONNumber": 5, "NUMBER": 6, "JSONNullLiteral": 7, "NULL": 8, "JSONBooleanLiteral": 9, "TRUE": 10, "FALSE": 11, "JSONText": 12, "JSONValue": 13, "EOF": 14, "JSONObject": 15, "JSONArray": 16, "{": 17, "}": 18, "JSONMemberList": 19, "JSONMember": 20, ":": 21, ",": 22, "[": 23, "]": 24, "JSONElementList": 25, "$accept": 0, "$end": 1 },
```
gandhiniraj: Any specific reason not to sort e.g. $accept : 0

lib/util/jsonLint.js
```diff
@@ -0,0 +1,440 @@
...
+        
+        // this shouldn't happen, unless resolve defaults are off
+        if (action[0] instanceof Array && action.length > 1) {
+          throw new Error('Parse Error: multiple actions possible at state: ' + state + ', token: ' + symbol);
+        }
+        
+        switch (action[0]) {
+
+          case 1:// shift
+            //this.shiftCount++;
```
gandhiniraj: remove?

lib/util/jsonLint.js
```diff
@@ -0,0 +1,440 @@
...
+
```
gandhiniraj: unwanted line

lib/util/jsonLint.js
```diff
@@ -0,0 +1,440 @@
...
+                recovering--;
+            } else { // error just occurred, resume old lookahead f/ before error
+              symbol = preErrorSymbol;
+              preErrorSymbol = null;
+            }
+            break;
+
+          case 2:// reduce
+            //this.reductionCount++;
+
```
gandhiniraj: unwanted line

lib/util/jsonLint.js
```diff
@@ -0,0 +1,440 @@
+﻿/* jshint ignore:start */
```
gandhiniraj: Unit tests?

package.json
```diff
@@ -84,6 +84,7 @@
     "ms-rest": "^1.12.0",
     "ms-rest-azure": "^1.12.0",
     "node-forge": "0.6.23",
+    "node-import": "0.9.2",
```
amarzavery: why do we need this. Can't this be solved by require statements? azure-cli is already heavy. We do not want to increase dependencies unnecessarily.

vivsriaus: I don't think it degrades the performance. I see a lot of third party packages that we already use for convenience. I felt that this was the easiest way to import external js files into node.

That said, I think there is a better/easier way to do this, instead of adding an external js file. My latest commit will import jsonlint module instead of using node-import and the external js file. Let me know if that looks good.

PR #2843, opened 2016-05-02
CLI: 'validator' package updated, improved common validation logic

romangromov:
_Motivation:_
It's common case to validate some cli input options/parameters, like:
*  `--ip-address should be valid ip address`, 
*  `--priotity should be valid Integer`, 
*  `--enable-forwarding should be Boolean`, and so on, routine..

Also we have already `validator` package included for such  a validation logic. But for some reason, it was not usefull and validation logic was duplicated in many and many places, until this PR.
Now `lib\util\validation.js` wraps `validator` package and provides one-place module for input options/parameters validation.
Pros: This will eliminate modules dependences, for example: no need to reference `lib\util\vnet.Util._js` to make some IP format validation, let `validator` do this!

@amarzavery @yugangw-msft please review, merge


lib/commands/asm/websites/validation.js
```diff
@@ -0,0 +1,88 @@
...
+  throw new Error('Required argument ' + name + ' for function ' + func + ' is not defined');
+}
+
+function ArgumentValidator(functionName) {
+  this.func = functionName;
+}
+
+_.extend(ArgumentValidator.prototype, {
+  string: function (val, name) {
+    if (typeof val != 'string' || val.length === 0) {
```
amarzavery: typeof val.valueOf() !== 'string'

lib/commands/asm/websites/validation.js
```diff
@@ -0,0 +1,88 @@
...
+  },
+
+  value: function (val, name) {
+    if (!val) {
+      throwMissingArgument(name, this.func);
+    }
+  },
+
+  nonEmptyArray: function (val, name) {
+    if (!val || val.length === 0) {
```
amarzavery: check whether val is an Array in the first place. Array.isArray(val)

lib/commands/asm/websites/validation.js
romangromov: @amarzavery thanks, done

lib/commands/asm/websites/validation.js
romangromov: @amarzavery thanks, done

PR #2821, opened 2016-04-26
Fix for buffer under node v6

xeoneux:
Fixes #2820 by removing buffer's encoding. Not required and defaults to `utf-8`

PR #2846, opened 2016-05-03
Remove encoding from storage util

mlrobinson:
* Similar to #2821, removing buffer's encoding is required to work with storage blobs under Node.js version 6.


azurecla:
Hi __@mlrobinson__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

PR #2849, opened 2016-05-03
Re-run VM Live Test

huangpf:


huangpf:
@amarzavery @haocs Test & record only. Could you please help review and merge it to the branch? This would update the latest test run status for VM related commands. Thanks.

PR #2883, opened 2016-05-16
Adding --applicationName to script-action create command for Edgenode…

nbikhchandani:
* 
* 
* 
… customization

Update Debugging.md

Updating indent

Updating indent

Updating indent and adding randomly generated appName

Update indent

Removing positive edgenode customization test

Generating recordings

Adding --applicationName to script-action create command for Edgenode customization

Update Debugging.md

Updating indent

Updating indent

Updating indent and adding randomly generated appName

Update indent

Removing positive edgenode customization test

Generating recordings

Updating indent

Update indent


nbikhchandani:
@amarzavery Hi Amar, the build failed for tests that my changeset doesn't touch at all. Could you please take a look? Looks like you are tracking the vm recording failures already. This seems to be the same error. Thanks.

lib/commands/arm/hdinsight/hdinsight._js
```diff
@@ -1003,6 +1003,7 @@ hdInsightCommandLine.init = function(cli) {
     .option('-u, --uri <uri>', $('Script URI'))
     .option('-p, --parameters <parameters>', $('(Optional) Script parameters'))
     .option('-t, --nodeTypes <nodeTypes>', $('Specifies the nodes on which this cmdlet applies the action separated by ";". For example, "headnode;workernode" is a valid input.'))
+    .option('-a, --applicationName <applicationName>', $('Application name. When application name is specified, persistOnSuccess cannot be specified, and nodeTypes must contain only "edgenode"'))
```
amarzavery: Was the validation logic for --persistOnSuccess and --aplicationName to be mutually exclusive already there in the CLI as I do not see that in this PR?

It would not be ideal to send a request and let the server fail.

nbikhchandani: This has recently been added to the RP code for the new Edgenode customization feature. The reason those two are being kept mutually exclusive is to avoid scenarios when application is uninstalled before the persisted script is demoted.

amarzavery: I am not reasoning "why they are mutually exclusive".
I am asking is the validation for mutual exclusiveness done in the CLI? If not, then do you think it would be better to do it in the CLI rather than sending the request to the server and then failing based on the response received by the server. Simple validation check by the CLI will fail faster.
What do you think?

amarzavery: Take a look at this example: https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/trafficManager._js#L378

nbikhchandani: That's fair. I'll add the validation. Thanks for pointing out!

PR #2890, opened 2016-05-19
Display provisioning status message during deployments

vivsriaus:
* Display provisioning status message during create deployment


lib/commands/arm/group/group.deployment._js
```diff
@@ -89,22 +95,16 @@ exports.init = function (cli) {
             var failedOperations = groupUtils.getFailedDeploymentOperations(options.subscription, resourceGroup, deployment.name, _);
 
             failedOperations.forEach(function (operation) {
+              //handle special case for web, where they send their own error message object
+              if (operation.properties.statusMessage.Message) {
```
amarzavery: just verifying:

is statusMessage a required property? If not then it is worthwhile to check if it exists before doing a .Message.

vivsriaus: @amarzavery Microsoft.Web is not following the ARM standard for error response content. From our API spec, we expect the error content to be like:

{ 
    “error”: { 
      "code": "BadArgument", 
      "message": "The provided database ‘foo’ has an invalid username.", 
      "target": "query", 
      "details": [ 
        { 
         "code": "301", 
         "target": "$search"   
         "message": "$search query option not supported", 
        } 
      ] 
      "innererror": { 
        "trace": [...], 
        "context": {...} 
      } 
    } 
} 

We'll need to follow up with them to fix this - but it's a good idea to check for statusMessage to be null either way. I've added that check in my last commit. Thanks for catching!

PR #2891, opened 2016-05-20
Add parameterized way to create/set pool and task

xingwu1:
* 
*  Add parameterized way to create/update batch entities
*


lib/commands/batch/batch.pool._js
```diff
@@ -60,7 +62,21 @@ exports.init = function(cli) {
     
   pool.command('create [json-file]')
     .description($('Create a Batch pool'))
-    .option('-f, --json-file <json-file>', $('the file containing the pool object to create in JSON format'))
+    .option('-f, --json-file <json-file>', $('the file containing the pool object to create in JSON format, if this parameter is specified, all other pool parameters are ignored'))
+    .option('-i, --id <pool-id>', $('the Batch pool id'))
+    .option('-s, --vm-size <vm-size>', $('the size of virtual machines in the pool, ex: small, Standard_D14'))
```
amarzavery: -s is reserved for subscription

xingwu1: We don't use subscription parameter for these data plane commands.

amarzavery: I see. However, we still treat -s as a special case for subscription and want to make sure that this is not used by any command apart from using it for subscription. That keeps things simple.

xingwu1: Fixed.

PR #2892, opened 2016-05-20
Add powerbiembedded cli commands with tests.

mattmazzola:
Add `azure powerbi *` cli commands.  Uses `azure-arm-powerbiembedded` package.

I'm not sure how to pusblish the `azure-arm-powerbiembedded` package from the azure-sdk-for-node using jenkins so I have the package json set to use local package here although the code is merged into the azure-sdk-for-node repo so you should be able to update it locally for testing.

Also, the tests don't run there is an error:
`errorText: 'error: callback missing (argument #3). See https://github.com/Sage/streamlinejs/blob/master/FAQ.md#no-callback-given-error\nerror`

I checked that every function passed to the cli.execute method has the last argument has an underscore for the node streamline to work properly.


azurecla:
Hi __@mattmazzola__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (mattm). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/commands/arm/powerbiembedded/powerbiembedded._js
```diff
@@ -0,0 +1,349 @@
...
+      /** Create client */
+      var subscription = profile.current.getSubscription(options.subscription);
+      var client = utils.createPowerbiManagementClient(subscription);
+      
+      /** Invoke client method */
+      var tags = {};
+      tags = tagUtils.buildTagsParameter(tags, options);
+    
+      var workspaceCollectionCreationOptions = {
+        location: options.location || "southcentral",
```
mattmazzola: Is it bad to use default values for location like this?  Should we force the user to specify a location?

lib/commands/arm/powerbiembedded/powerbiembedded._js
```diff
@@ -0,0 +1,349 @@
...
+      var progress = cli.interaction.progress(util.format($('Udating workspace collection: %s'), name));
+      var workspaceCollection;
+      try {
+        workspaceCollection = client.workspaceCollections.update(resourceGroup, name, body, _);
+      } finally {
+        progress.end();
+      }
+      
+      /** Display output */
+      log.info('WorkspaceCollection updated!');
```
mattmazzola: Do we have to use the localized `$(...)` helper for log strings?

lib/commands/arm/powerbiembedded/powerbiembedded._js
amarzavery: What if the user is in China Cloud or hybrid cloud or German Government cloud. The default location will not work. It is always better to ask the user to provide the location.

What if the latest api-version has not been provisioned in southcentral?

lib/commands/arm/powerbiembedded/powerbiembedded._js
```diff
@@ -0,0 +1,349 @@
...
+        sku: {
+          name: "S1",
+          tier: "Standard"
+        }
+      };
+
+      var progress = cli.interaction.progress(util.format($('Creating workspace collection: %s'), name));
+      var workspaceCollection;
+      try {
+        console.log('creation parameters: ', name, workspaceCollectionCreationOptions);
```
amarzavery: Please do not use console.log. Please use log.info or log.verbose

amarzavery: Also make sure that the output is correct with --json switch.

lib/commands/arm/powerbiembedded/powerbiembedded._js
amarzavery: Yup

lib/commands/arm/powerbiembedded/powerbiembedded._js
```diff
@@ -0,0 +1,349 @@
...
+    .option('-s --subscription [subscription]', $('The subscription identifier'))
+    .execute(function (options, _) {
+      /** Create client */
+      var subscription = profile.current.getSubscription(options.subscription);
+      var client = utils.createPowerbiManagementClient(subscription);
+      
+      /** Invoke client method */
+      var workspaceCollections;
+      if(options.resourceGroup) {
+        var progress = cli.interaction.progress($('Getting workspace collections in subscription ' + subscription + ' resource group: ' + options.resourceGroup));
```
amarzavery: Usually customers work in a particular subscription. They always set the subscription by `azure account set "subname"` Hence they know the subscription. So showing the subscription is not useful. 

Btw, that is a full blown subscription object with lot of other info. It has a property named id. Basically it is the lib/util/profile/subscription.js

lib/commands/arm/powerbiembedded/powerbiembedded._js
amarzavery: That is how we usually display the output. https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/providers/provider._js#L44 Take a look at group list, show, create commands for examples.

lib/commands/arm/powerbiembedded/powerbiembedded._js
```diff
@@ -0,0 +1,349 @@
...
+      /** Validate parameters */
+      if (!resourceGroup) {
+        return cli.missingArgument('resourceGroup');
+      }
+      if (!name) {
+        return cli.missingArgument('name');
+      }
+      if (!keyName) {
+        return cli.missingArgument('keyName');
+      }
```
amarzavery: Do you want to validate the keyName value? `keyName.match/^key(1|2)$/ig `

lib/commands/arm/powerbiembedded/powerbiembeddedUtils._js
```diff
@@ -0,0 +1,22 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+var utils = require('../../../util/utils');
+
+exports.createPowerbiManagementClient = function (subscription) {
```
amarzavery: Sorry for the wrong info in the template. We shall update the sample. 
Please take a look at this example https://github.com/Azure/azure-xplat-cli/blob/dev/lib/util/utils.js#L233 and add this in the lib/util/utils.js as other services have.

amarzavery: You actually don't need this file at all.

package.json
```diff
@@ -127,6 +128,7 @@
     "unit-arm": "node scripts/unit.js testlist-arm.txt",
     "unit-vm": "node scripts/unit.js testlist-vm.txt",
     "unit-vm-live": "node scripts/unit.js testlist-vm-live.txt",
+    "unit-arm-powerbiembedded": "node scripts/unit.js testlist-arm-powerbiembedded.txt",
```
amarzavery: this pattern is usually used by compute as they are very big RP and have lot of services underneath it.In you case, you can just add your tests to testlist-arm.txt and you should be good.

test/testlist-arm-powerbiembedded.txt
```diff
@@ -0,0 +1,2 @@
+#owner: @mattm
```
amarzavery: this file is not required

lib/commands/arm/powerbiembedded/powerbiembedded._js
```diff
@@ -311,10 +311,10 @@ exports.init = function (cli) {
...
     });
     
-    /**
-     * List workspaces within workspace collection
-     */
-    powerbiembeddedcli.command('list-workspaces <resourceGroup> <name>')
+  /**
+   * List workspaces within workspace collection
+   */
+  powerbiembeddedcli.command('list-workspaces <resourceGroup> <name>')
```
amarzavery: we follow the noun verb pattern.
create a new category called workspaces and you should have list method in it.

`azure powerbi workspaces list`

lib/commands/arm/powerbiembedded/powerbiembedded._js
mattmazzola: Yea, I will change this.  Originally the powerbi service was only in south central so it didn't make sense to ask for location since it was constant but we just opened it to have all regions so it makes sense to make it required parameter now.

lib/commands/arm/powerbiembedded/powerbiembedded._js
mattmazzola: Oh ok, I see why the formatOutput is needed now. Will change them.

lib/commands/arm/powerbiembedded/powerbiembedded._js
mattmazzola: Ok, I will add the `id` property.  I wanted to make it clear that this only shows the workspace collections that are the intersection of subscription and resource group.

Subscriptions and Resource groups can each have multiple workspace collections but the only api we support is getting workspace collectiosn within a resource group *and* also within subscription so it can be misleading.

lib/commands/arm/powerbiembedded/powerbiembedded._js
mattmazzola: Yea, I will do that. Our api would catch it but it's better to fail faster i think

package.json
mattmazzola: Oh ok, will remove then. Thanks

test/testlist-arm-powerbiembedded.txt
mattmazzola: Ok, will remove

mattmazzola:
Is this good to merge or are there any other changes I need to make?

I was hoping this could get in for the next release.

lib/commands/arm/powerbiembedded/powerbiembedded._js
```diff
@@ -0,0 +1,422 @@
...
+          row.cell($('Provisioning State'), workspaceCollection.properties.provisioningState);
+          row.cell($('Tags'), tagUtils.getTagsInfo(workspaceCollection.tags));
+        });
+      });
+    });
+
+  /**
+   * Get access keys for workspace collection
+   */
+  powerbiembeddedcli.command('get-keys <resourceGroup> <name>')
```
amarzavery: This should keys list. Add a category named keys and the comman name will be list
storage also has a similar command
`azure storage account keys list`

lib/commands/arm/powerbiembedded/powerbiembedded._js
```diff
@@ -0,0 +1,422 @@
...
+          row.cell($('Key'), key.key);
+        });
+      });
+    });
+  
+  /**
+   * Regenerage access key
+   */
+  powerbiembeddedcli.command('regenerate-key <resourceGroup> <name> <keyName>')
+    .description($('Get access keys for a workspace collection'))
```
amarzavery:
```
D:\sdk\azure-sdk-for-node\lib\services\storageManagement2\lib>azure storage account -h
help:    Commands to manage your Storage accounts
help:
help:    Commands to manage your Storage account keys
help:      storage account keys list [options] <name>
help:      storage account keys renew [options] <name>
help:
```
Please follow the same format wrt to structure and naming for consistency.

lib/commands/arm/powerbiembedded/powerbiembedded._js
```diff
@@ -0,0 +1,422 @@
...
+    .option('-g --resource-group <resourceGroup>', $('Name of the resource group'))
+    .option('-n --name <name>', $('Name of workspace collection'))
+    .option('-k --keyName <keyName>', $('Name of key you would like to regenerate. Accepted values: "key1" (Primary) or "key2" (Secondary)'))
+    .option('-s --subscription [subscription]', $('The subscription identifier'))
```
amarzavery:
```
D:\sdk\azure-sdk-for-node\lib\services\storageManagement2\lib>azure storage account keys renew -h
help:    Renew a key for a storage account from your account
help:
help:    Usage: storage account keys renew [options] <name>
help:
help:    Options:
help:      -h, --help                            output usage information
help:      -v, --verbose                         use verbose output
help:      -vv                                   more verbose with debug output
help:      --json                                use json output
help:      --primary                             Update the primary key
help:      --secondary                           Update the secondary key
help:      -g, --resource-group <resourceGroup>  the resource group name
help:      -s, --subscription <id>               the subscription id
help:
help:    Current Mode: arm (Azure Resource Management)

D:\sdk\azure-sdk-for-node\lib\services\storageManagement2\lib>
```

lib/commands/arm/powerbiembedded/powerbiembedded._js
```diff
@@ -0,0 +1,422 @@
...
+          row.cell($('Name'), key.name);
+          row.cell($('Key'), key.key);
+        });
+      });
+    });
+    
+  /**
+   * List workspaces within workspace collection
+   */
+  powerbiembeddedcli.command('list-workspaces <resourceGroup> <name>')
```
amarzavery: workspaces list.

https://github.com/Azure/azure-xplat-cli/blob/dev/Documentation/Writing-Cmd.md#command-design-guidelines

amarzavery:
@mattmazzola  the CI is failing due to this reason https://travis-ci.org/Azure/azure-xplat-cli/jobs/132659697#L571

test/commands/arm/powerbiembedded/arm.powerbiembedded-tests.js
```diff
@@ -0,0 +1,161 @@
...
+      suite = new CLITest(this, testPrefix, requiredEnvironment);
+      
+      // set defaults from environment variables
+      accountLcation = process.env.AZURE_POWERBIEMBEDDED_TEST_LOCATION || "southcentralus";
+      accountSku = process.env.AZURE_POWERBIEMBEDDED_TEST_SKU;
+      accountTier = process.env.AZURE_POWERBIEMBEDDED_TEST_TIER;
+      
+      suite.setupSuite(function () {
+        resourceGroupName = suite.generateId(resourceGroupPrefix, createdResourceGroups);
+        suite.execute('group create %s %s --json', resourceGroupName, accountLcation, function (result) {
```
amarzavery: We dont record the commands that are executed in before and after as they are not required. What is required is the name of the artifacts created in this section. In this case the name of the resource group.
the suite.execute must be wrapped in an if condition that runs the command if the mode is not playback
```
suite.setupSuite(function () {
  resourceGroupName = suite.generateId(resourceGroupPrefix, createdResourceGroups);
  if (!suite.isPlayback()) {
    suite.execute('group create %s %s --json', resourceGroupName, accountLcation, function (result) {
      result.exitStatus.should.equal(0);
      done();
    });
  } else {
    done();
  }
});
```

Same thing in `after()` as well.
We documented this in [best practices for reducing playback time](https://github.com/Azure/azure-xplat-cli/blob/dev/Documentation/BestPractices-Testing.md#ways-to-reduce-playback-time)but have not documented this in the main doc "How to write a test". Sorry about that.

amarzavery: You just need to modify this and then when you run the tests in playback mode on your machine, it should just work.
You dont need to record tests again.

mattmazzola:
@amarzavery Thanks for you help! I applied the isPlayback changes you suggested and it's working now.

PR #2894, opened 2016-05-23
CPlat PR : Fix BGInfo's major version search issue

huangpf:


lib/commands/arm/vm/vmClient._js
```diff
@@ -179,12 +184,12 @@ __.extend(VMClient.prototype, {
...
 
     // Enable BGInfo for Windows VM if found
-    if (found) {
+    if (foundBginfoExt) {
       var bgInfoParams = {
         resourceGroupName: resourceGroupName,
         osType: osType,
         location: location,
-        version: foundVersion
+        version: foundMajorVersion
```
yugangw-msft: If i understand it right, the bgInfoParams.version is now being used for 2 purposes, a boolean flag and a string with the real version. If yes, then i would suggest a new field called "versionFound", and also the code set the real "version" right here. This should help the readability

huangpf: OK. Thanks.

huangpf:
@amarzavery Could you please help review? Thanks.

PR #2902, opened 2016-05-25
CPlat : VMSS Scaleout Command

huangpf:


test/commands/arm/vmss/arm.vmss-create-tests.js
```diff
@@ -230,6 +230,26 @@ describe('arm', function() {
           done();
         });
       });
+            
+      it.skip('vmss scale up should pass', function (done) {
```
amarzavery: @huangpf - Why are the tests skipped?

huangpf: Dear @amarzavery, I'm still trying to get the tests recorded. As it takes long time to record, & redo everytime if it fails... Please help wait a bit, thanks...

PR #2903, opened 2016-05-25
expose a switch to use file based token storage

yugangw-msft:


lib/util/authentication/adalAuth.js
```diff
@@ -25,10 +25,12 @@ var adal = require('adal-node');
 var utils = require('../utils');
 var defaultTokenCacheFile = path.join(utils.azureDir(), 'accessTokens.json');
 
+var useSecureTokenStorage = utils.ignoreCaseEquals(process.env['AZURE_USE_SECURE_TOKEN_STORAGE'], 'true') ||
```
amarzavery: @yugangw-msft  - why can't we keep this simple as follows?
```
var useSecureTokenStorage = false;
if (process.env['AZURE_USE_SECURE_TOKEN_STORAGE']) useSecureTokenStorage = true;
```
I don't care if the user set's it true, True, Foo, 1. If it is a truthy value then we use the secure tokenCache. If it is a falsy value (0, undefined)then we don't.

yugangw-msft: There is no truthy values, it is either set or not set unless we parse the value
For now till the SDLC gets approved. People must opt-in the file storage explicitly. With you proposed change, the file based storage becomes the default, which is not what we want.
This flag, for most people, should be completelytransparent. With that, i like people who set the flag will see no behavior change once we switch the default value.

PR #2911, opened 2016-05-31
Remove custom 409 error handling from powerbi create command since it's not always accurate.

mattmazzola:
The 409 can mean either name conflict with existing workspace collection or also be returned due to the subscription not having the Microsoft.Powerbi RP registered.  Removing the custom error handling and allowing the original error to be thrown will ensure proper response body / error message is displayed.


mattmazzola:
@amarzavery This pull request is response to your email about wrong error message being printed.  I think this change will prevent the wrong error message from being returned since no longer attempt to cover it with a friendly message.

haocs:
@mattmazzola - could you please target your PR to release-0.10.1 branch since it will be included in this release.

mattmazzola:
Ok, closing this and re-opening as #2912

PR #2912, opened 2016-05-31
Remove custom 409 error handling from powerbi create command since it's not always accurate.

mattmazzola:
The 409 can mean either name conflict with existing workspace collection or also be returned due to the subscription not having the Microsoft.Powerbi RP registered. Removing the custom error handling and allowing the original error to be thrown will ensure proper response body / error message is displayed.


mattmazzola:
Duplicate of #2911 with target branch changed to release-0.10.1

PR #2913, opened 2016-06-01
Fix bug in batch account update command

xingwu1:
* 
* 
*


haocs:
@xingwu1 - could you target your PR to release-0.10.1 branch since it will be included in this release? thanks

PR #2914, opened 2016-06-01
Fix bug in batch account update command.

xingwu1:
* 
* 
*


xingwu1:
@haocs resend PR at correct branch.

PR #2922, opened 2016-06-03
CLI: table output improved

romangromov:
This improvement should allow to avoid constructions like `row.cell('cell', value || '');` as `''` will be substituted automatically

@amarzavery @yugangw-msft please merge

### DIFF
--- a/lib/util/logging.js
+++ b/lib/util/logging.js
@@ -224,6 +224,14 @@ log.table = function (level, data, transform) {
       }
     }
 
+    for (var i in table.lines) {
+      for (var column in table.columns) {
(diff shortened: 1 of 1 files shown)


haocs:
@romangromov - Looks like it will traverse each table.line and scan all column property against that line. Will it potentially cause a performance issue?

romangromov:
This change is required for CLI automatic generator as currently (in the manually written CLI code) required fields names are hardcoded and every field value that can be potentially `undefined` is added via `||` operator.
Generator needs to iterate over all object fields recursively and it is impossible to define that field value should be replaced with an empty string before all the table columns are known.

Issue #2767, opened 2016-04-15
vm extension set: No way to specify autoUpgradeMinorVersion


ahmetalpbalkan:
CLI Version: **0.9.19**
Mode: **ARM**

ASM mode had a feature to let users pick the highest minor version of a VM extension through version syntax, such as `1.*`. So if I run:

    [asm] azure vm extension set [...] Extension Namespace 1.*

I will get the highest `minor` version with `major=1`. This feature was already there in ASM because it's a feature of Service Management APIs and has nothing to do with client-side.

Now, in ARM, there is no way of doing this and this is actually a pretty useful/important feature we should be encouraging users to follow. ARM does not accept `1.*` syntax anymore, but in the extension properties it accepts an object:

    "autoUpgradeMinorVersion": true,

With that, users can pick the highest version and just type `1.0` but a version like `1.5` will get installed to their VM. To make this happen we need a `--autoUpgradeMinorVersion` argument that would set the respective object key in request JSON to true.

Don't ever worry what autoUpgradeMinorVersion means, it is a long story and it doesn't mean an upgrade.

sauryadas:
@huangpf Is this something available in the API that we could set or does the service API need to be modified to allow for a autoUpgradeMinorVersion parameter?

ahmetalpbalkan:
@sauryadas it exists in the API already, but no way to set it in the CLI, that's the issue.

ahmetalpbalkan:
folks, any updates? this should be very easy to add.

yugangw-msft:
@huangpf, I assume your own the command? This field is already circulated in the help file, and you just need to expose the option.

ahmetalpbalkan:
@huangpf ping.

huangpf:
I think we shall expose this parameter for users. + @sauryadas @sharadagarwal FYI

huangpf:
![tumblr_nncf71ayoa1qaorn7o1_r4_500 1](https://cloud.githubusercontent.com/assets/3858524/16087151/db7f89ca-32d6-11e6-9c23-97e97a43ee3d.gif)

PR #2943, opened 2016-06-17
Exposed  --autoUpgradeMinorVersion  for vm extension. Resolved issue  #2767

shraddhaagrawal:
@huangpf and @sauryadas Please check


lib/commands/arm/vm/vmClient._js
```diff
@@ -2070,3 +2070,4 @@ __.extend(VMClient.prototype, {
 
 
 module.exports = VMClient;
+
```
huangpf: This file's change shall be reverted.

lib/commands/arm/vm/vmExtensionProfile._js
```diff
@@ -913,6 +913,7 @@ __.extend(VMExtensionProfile.prototype, {
       virtualMachineExtensionType: params.extensionName,
       typeHandlerVersion: params.version,
       settings: null,
+	  autoUpgradeMinorVersion: params.autoUpgradeMinorVersion,
```
huangpf: Alignment problem. Please use editor that can view and edit spaces.

huangpf:
Please add test case. Then it looks good to me. Thanks.

huangpf:
Should fix #2767

lib/commands/arm/vm/vmExtensionProfile._js
huangpf: If this parameter is not specified, is it ```false```, ```null```, or ```undefined```?

The code logic should be:
1) not specify => null
2) specify  => true

Please refer to: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators

shraddhaagrawal:
Thanks @huangpf  
closed following comments-
1.Alignment Issue
2.If parameter is not specified then it is null.
3.Added test cases and recorded files for record mode.
4.Revert changes in vmClient file.
Please review.

lib/commands/arm/vm/vmExtensionProfile._js
```diff
@@ -913,6 +913,7 @@ __.extend(VMExtensionProfile.prototype, {
       virtualMachineExtensionType: params.extensionName,
       typeHandlerVersion: params.version,
       settings: null,
+      autoUpgradeMinorVersion: null,
```
huangpf: @shraddhaagrawal 
You should evaluate the ```options.autoUpgradeMinorVersion```. I think currently it's always set to ```null```?

test/commands/arm/vm/arm.vm.extension-tests.js
```diff
@@ -178,6 +179,28 @@ describe('arm', function() {
...
+        });
+      });
+	  
+	  it('Extension Get should should give autoUpgradeMinorVersion=true ', function(done) {
+        this.timeout(vmTest.timeoutLarge);
+        var cmd = util.format('vm extension get %s %s --json', groupName, vmPrefix).split(' ');
+        testUtils.executeCommand(suite, retry, cmd, function(result) {
+          result.exitStatus.should.equal(0);
+          var allResources = JSON.parse(result.text);
+		  allResources[1].autoUpgradeMinorVersion.should.be.true;
```
huangpf: alignment...

lib/commands/arm/vm/vmExtensionProfile._js
shraddhaagrawal: @huangpf  Please check line 951 where code is evaluating params.autoUpgradeMinorVersion and setting value of autoUpgradeMinorVersion = true.

huangpf: OK, interesting. I guess other extension commands will actually set this parameter and when it comes to this common code path, it will take if it’s set. And it’s just now for our generic extension command to expose it.

The PR looks good to me, and should be ready to merge.

huangpf:
@amarzavery @haocs Could you please help review the fix for #2767? Thanks

amarzavery:
LGTM

lib/commands/arm/vm/vm._js
```diff
@@ -586,6 +586,7 @@ exports.init = function (cli) {
       .option('-t, --tags <tags>', $('Tags to set to the resource group. Can be mutliple. ' +
               'In the format of \'name=value\'. Name is required and value is optional. For example, -t tag1=value1;tag2'))
       .option('-q, --quiet', $('quiet mode, do not ask for uninstall confirmation'))
+      .option('-U, --auto-upgrade-minor-version', $('Enable auto-upgrade across minor version'))
```
ahmetalpbalkan: Documentation is not good here. It means:

    Pick the highest minor version of the extension

amarzavery: @ahmetalpbalkan - I do not have full context on this scenario. Can you please suggest better help text?
/cc @huangpf @sharadagarwal

ahmetalpbalkan: @amarzavery I already did above. Unfortunately the name is quite misleading, I don't know why CRP folks picked such a name. This is the equivalent of the `1.*` notation before, i.e. it picks the highest minor version of the specified major version during the deployment time. Only the hotfixes are applied to the extension afterwards. So we never auto-upgrade to new minor versions...

huangpf: It actually looks fine to me, as least it's not a blocking issue. But we have other PRs to submit for this sprint. I would suggest we change them altogether by then.

@sharadagarwal In the next few PRs, please inquire @sauryadas for the best help texts for updated command/parameters, as he owns this part. Thanks.

ahmetalpbalkan: not sure how it looks fine to you. it is 100% inaccurate

PR #2945, opened 2016-06-18
Azure CPlat PR -- Fix VMSS QC & VM Tests

huangpf:


huangpf:
@amarzavery @haocs Could you please help review? Thanks.

lib/commands/arm/vm/vm._js
```diff
@@ -90,6 +90,7 @@ exports.init = function (cli) {
     .option('-k, --vnet-subnet-address-prefix <vnet-subnet-address-prefix>', $('the virtual network subnet address prefix in IPv4/CIDR format'))
     .option('-r, --availset-name <availset-name>', $('the availability set name'))
     .option('-o, --storage-account-name <storage-account-name>', $('the storage account name'))
+    .option('--storage-account-type <storage-account-type>', $('the storage account type for to-be-generated storage account, default Premium_LRS'))
```
amarzavery: There is no account type with the latest version of azure-arm-storage library. Are you sure you want to do this?

amarzavery: There was breaking change in their API, in the way storage accounts are created. Please take a look: https://github.com/Azure/azure-rest-api-specs/blob/master/arm-storage/2016-01-01/swagger/storage.json#L574

huangpf: I think I already updated the creation parameter accordingly:
```
    stoParams.sku = { name: storageType };
    stoParams.kind = 'Storage';
```

amarzavery: Ah I see. Thanks

Issue #2939, opened 2016-06-15
azure account import errors with "Invalid IV length"


molant:
CLI Version: **0.10.1**
OS Type: Win
Installation via:  npm
node version: 6.2.1

Mode: **ARM / ASM**

Environment: **AzureCloud/Azure China Cloud/US Government/Blackforest/Azure Stack**

Description:
`azure account import credentials.publishsettings` command throws the following:
> info:    Executing command account import
error:   Invalid IV length
info:    Error information has been recorded to C:\Users\molant\.azure\azure.err
error:   account import command failed

Steps to reproduce:
1) Run `azure account import credentials.publishsettings`

Error stack trace:
```
2016-06-15T20:10:38.264Z:
{ Error: Invalid IV length
  <<< async stack >>>
  <<< raw stack >>>
    at Error (native)
    at new Decipheriv (crypto.js:232:16)
    at Object.Decipheriv (crypto.js:229:12)
    at Array.__dirname.PfxAlgorithmDecoder.self.on.elements.createCipher (C:\Users\molant\AppData\Roaming\npm\node_modules\azure-cli\lib\util\certificates\pkcs.js:274:23)
    at PfxPkcs8ShroudedKeyBagDecoder.<anonymous> (C:\Users\molant\AppData\Roaming\npm\node_modules\azure-cli\lib\util\certificates\pkcs.js:333:28)
    at emitOne (events.js:96:13)
    at PfxPkcs8ShroudedKeyBagDecoder.emit (events.js:188:7)
    at PfxPkcs8ShroudedKeyBagDecoder.Decoder.parse (C:\Users\molant\AppData\Roaming\npm\node_modules\azure-cli\lib\util\certificates\der.js:58:8)
    at PfxSafeBagDecoder.<anonymous> (C:\Users\molant\AppData\Roaming\npm\node_modules\azure-cli\lib\util\certificates\pkcs.js:305:27)
    at emitOne (events.js:96:13)
  stack: [Getter/Setter],
  __frame: undefined,
  rawStack: [Getter] }
Error: Invalid IV length
  <<< async stack >>>
  <<< raw stack >>>
    at Error (native)
    at new Decipheriv (crypto.js:232:16)
    at Object.Decipheriv (crypto.js:229:12)
[...]

PR #2964, opened 2016-06-27
Use buffer in place of binary string since crypto use 'utf8' as default string encoding since NodeV6

haocs:
* Fixed issue https://github.com/Azure/azure-xplat-cli/issues/2939
* 
*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 
 node_js:
-  - "6.0"
   - "4.2.4"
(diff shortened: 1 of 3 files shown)


haocs:
@yugangw-msft - Please help review this PR. I have tested 'account import' as well as 'account cert export' against NodeV6 and V4.

yugangw-msft:
LGTM, please revert the unrelated change on the CDN tests

PR #2971, opened 2016-06-29
Resolved issue with ARM VM Show method

MikhailTryakhov:
* Changed show vm functionalirt: throw error instead of warning if incorrect name of VM was entered

### DIFF
--- a/lib/commands/arm/vm/vmClient._js
+++ b/lib/commands/arm/vm/vmClient._js
@@ -378,7 +378,7 @@ __.extend(VMClient.prototype, {
       if (isJson) {
         output.json({});
       } else {
-        output.warn($('No VMs found'));
+        throw new Error($('No VMs found'));
       }
     }
   },


azurecla:

TTYL, AZPRBOT;

haocs:
@MikhailTryakhov - This PR looks good. It will be great if you could update the 'vm show' testing accordingly as well.

MikhailTryakhov:
HI @haocs 
will discuss about it - assume we should work on mock tests a lot because they're too simple and can not handle a lot of potentially problem bugs.
For example, a lot of associated scripts can fail after simple property renaming. No tests can handle it now.

haocs:
@MikhailTryakhov - Yes, totally agree. We can do this after this release(next Thursday). I assume there will be many changes.

MikhailTryakhov:
@haocs can we sync up about test cases which can help us close most of possible issues?
- Test incorrect values ('it should not pass' tests)
- Try to find items, which don't exist
- Test duplicated items

What do you think, which other cases can be helpful in tests? 
Feel free to ping me to make a call

Issue #2961, opened 2016-06-25
VM list command doesn't follow continuation tokens when not supplied RG


AlanSt:
CLI Version: **0.10.0**
OS Type: Linux (Ubuntu)
Installation via:  npm 

Mode: **ARM**

Environment: **AzureCloud**

Description:
`azure vm list` command doesn't output follow continuation tokens when not supplied resource group

huangpf:
By 'continuation tokens', do you mean pagination via nextLink in the response?

huangpf:
How many VMs? I can't repro by 51. Please share the ```-vv``` logs. Thanks.

AlanSt:
Only lists first 50 per region, additional VMs cannot be listed through this method.  Will send you logs.

huangpf:
Got it. For listAll() function it doesn't work for pagination to include all VMs from various pages, but it seems list(rgName) works just fine, as there is no pagination.

```
],
  "nextLink": "https://management.azure.com:443/subscriptions/xxxxxxxxxxxxxxxxxxxxxxxx/providers/Microsoft.Compute/virtualMachines?api-version=2016-03-30&$skiptoken=......................"
}
```

huangpf:
@amarzavery The Swagger spec specifies ```nextLink``` but Node.js SDK doesn't deserialize this value. Could you please help look into it?

Ref:
https://github.com/Azure/azure-rest-api-specs/blob/master/arm-compute/2016-03-30/swagger/compute.json#L1202

Ref:
https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/computeManagement2/lib/models/virtualMachineListResult.js

huangpf:
Will be fixed in the new sprint with ```0.19.0``` upgrade of the underlying compute package.

PR #2976, opened 2016-06-30
Fix VM List/ListAll #2961

huangpf:
* Fix VM List/ListAll #2961


huangpf:
@haocs Please help review.

PR #2982, opened 2016-07-01
azure keyvault: added options to upload and download secrets as a file

msfcolombo:
Text to be added to changelog:
* Key Vault
  * Added options to upload a file as a secret to `azure keyvault secret set`.
  * Added new command `azure keyvault secret get` to download a secret as a file.
  * The `azure keyvault secret` commands `set`, `set-attributes` and `delete` don't echo the secret contents by default anymore.


azurecla:
Hi __@msfcolombo__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@msfcolombo__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

lib/commands/arm/keyvault/keyvault.secret._js
```diff
@@ -313,16 +327,55 @@ exports.init = function(cli) {
...
+        secret = client.secrets.get(secretIdentifier, _);
+      } finally {
+        progress.end();
+      }
+
+      options.passThru = true;
+      showSecret(options, secret);
+    });
+
+  secret.command('get [vault-name] [secret-name] [secret-version] [file]')
```
amarzavery: @msfcolombo - Please use the approved verb show for consistency instead of get.

amarzavery: Actually how about download as the command name. Makes it more clear to the user.

lib/commands/arm/keyvault/keyvault.secret._js
```diff
@@ -313,16 +327,55 @@ exports.init = function(cli) {
...
+    .description($('Downloads a secret from the vault'))
+    .usage('[--vault-name] <vault-name> [--secret-name] <secret-name> [[--secret-version] <secret-version>] [--file] <file-name> [options]')
+    .option('-u, --vault-name <vault-name>', $('the vault name'))
+    .option('-s, --secret-name <secret-name>', $('the secret name'))
+    .option('-r, --secret-version <secret-version>', $('the secret version; if ommited, uses the most recent'))
```
amarzavery: What could be the possible values for the secret version. Some examples or a link to an MSDN REST API doc that explains it with sample values would be useful

lib/commands/arm/keyvault/keyvault.secret._js
msfcolombo: The command `show` already exists and does a different thing. The `get` verb is more consistent with another correspondent command, `set`.

lib/commands/arm/keyvault/keyvault.secret._js
msfcolombo: From my support experience, most KV users understand the secret version. It's hard to specify an example, because it's a random hex string defined by the service when the secret is created or updated. The `set` command outputs the version when a secret is created.

PR #3012, opened 2016-07-14
Reworked app gateway long-running commands.

MikhailTryakhov:
* Added option to force app gateway modification commands to avoid long waiting for operations finish.


azurecla:

TTYL, AZPRBOT;

lib/commands/arm/network/appGateway._js
```diff
@@ -273,6 +273,13 @@ __.extend(AppGateways.prototype, {
 
     var progress = self.interaction.progress(util.format($('Starting an application gateway "%s"'), appGatewayName));
     try {
+      if(options.forceStop) {
```
amarzavery: instead of doing this, please take a look at the --nowait implementation for [`azure group delete`](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/group/group._js#L144).

When the user passes --nowait switch the CLI should directly call the begin* method. This will send the request and provide the initial response 201/202. The user can then do a show on the resource to find out the status.

amarzavery: also use the --nowait as the switch. That will be consistent with other commands.

lib/commands/arm/network/appGateway._js
```diff
@@ -289,7 +296,7 @@ __.extend(AppGateways.prototype, {
 
     var progress = self.interaction.progress(util.format($('Stopping an application gateway "%s"'), appGatewayName));
     try {
-      self.networkManagementClient.applicationGateways.stop(resourceGroup, appGatewayName, _);
+      self.networkManagementClient.applicationGateways.stop(resourceGroup, appGatewayName, null, _);
```
amarzavery: Is there a reason for this to be null? May be that is why you see the following error, as we usually do options.subscription.
```
Uncaught ReferenceError: subscription is not defined
```

amarzavery:
having atleast one test with --nowait would be useful.

MikhailTryakhov:
thanks @amarzavery 
updated functionality. Hope tests would work fine

test/commands/arm/network/arm.network.application-gateway-tests.js
```diff
@@ -582,6 +582,52 @@ describe('arm', function () {
...
+          networkUtil.shouldHaveTags(appGateway);
+          networkUtil.shouldBeSucceeded(appGateway);
+          done();
+        });
+      });
+
+      it('delete should delete application gateway without waiting', function (done) {
+        var cmd = 'network application-gateway delete {group} {name} -q --nowait --json'.formatArgs(gatewayProp);
+        testUtils.executeCommand(suite, retry, cmd, function (deleteResult) {
+          console.log('DELETE result %j',deleteResult);
```
amarzavery: remove the console.log statement

MikhailTryakhov:
@amarzavery sorry for that. If fixed

PR #3033, opened 2016-07-22
Update for Storage commands

hasonmsft:

* Upgraded the azure-storage dependency to version 1.1.0.

* Fixed the issue that downloading public blob with size greater than 32MB will fail.

* Improved the error message for invalid account name when creating storage account.

* Added `--concurrenttaskcount` option for `azure storage file download` command. 

* Changed the default blob type to page blob when uploading VHD files.

* Added `--snapshot` option to `azure storage blob show` and `azure storage blob download` commands.

* Added `--lease` option for `azure storage blob upload` command to support overwrite a existing blob which is already leased.


azurecla:
Hi __@hasonmsft__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

PR #3047, opened 2016-07-28
[Shoebox] Support for event hub in diagnosticsettings

vcanaa:


vcanaa:
I am looking at the broken tests

PR #3054, opened 2016-07-29
[not for merge] ARM vnet peering implemented (autogenerated)

MikhailTryakhov:
* autogenerated ARM vnet peerings implemented


azurecla:

TTYL, AZPRBOT;

EvgenyAgafonchikov:
Tests are passing with new SDK however manual call of create/set/show fails: `error:   utils is not defined`
(1) Please include `utils` into generatorUtils.js

After fixing of (1) there is the following error inside of output:
```
error:   Object #<Object> has no method 'traverse'
FATAL ERROR: JS Allocation failed - process out of memory
```
(2) Please fix usage of `self` into generatorUtils.js

lib/commands/arm/network/virtualNetworkPeerings._js
```diff
@@ -0,0 +1,355 @@
...
+
+
+  virtualNetworkPeerings.command('create [resource-group] [vnet-name] [name]')
+    .description($('Create a virtual network peering'))
+    .usage('[options] <resource-group> <vnet-name> <name>')
+    .option('-g, --resource-group <resource-group>', $('the resource-group of virtual network peering'))
+    .option('-e, --vnet-name <vnet-name>', $('the virtual-network-name of virtual network peering'))
+    .option('-n, --name <name>', $('the virtual-network-peering-name of virtual network peering'))
+    .option('--remote-vnet-id <remote-vnet-id>', $('the remote-virtual-network-id of virtual network peering'))
+    .option('--allow-forwarded-traffic <allow-forwarded-traffic>', $('the allow-forwarded-traffic of virtual network peering'))
```
amarzavery: help for all the switches from here onwards for this command does not make sense.
For example:
--allow-forwarded-traffic <allow-forwarded-traffic> 'describes whether to allow traffic forwarding or not. Valid values are true or false'

Infact such boolean switches should just be flags. That is they should not be accepting any values. If specified then the value is true else false. Something like --enable

lib/commands/arm/network/virtualNetworkPeerings._js
```diff
@@ -0,0 +1,355 @@
...
+      }
+
+      var parameters = {};
+      if(options.remoteVnetId) {
+        parameters.remoteVirtualNetwork = {};
+        parameters.remoteVirtualNetwork.id = options.remoteVnetId;
+
+      }
+      if(options.allowForwardedTraffic) {
+        parameters.allowForwardedTraffic = utils.parseBool(options.allowForwardedTraffic);
```
amarzavery: the help does not say what are the set of valid values for this command. true, false, yes, no, y, n. If you use them as boolean flags then you don't need to do this.

lib/commands/arm/network/virtualNetworkPeerings._js
```diff
@@ -0,0 +1,355 @@
...
+    .option('-e, --vnet-name <vnet-name>', $('the virtual-network-name of virtual network peering'))
+    .option('-n, --name <name>', $('the virtual-network-peering-name of virtual network peering'))
+    .option('--remote-vnet-id <remote-vnet-id>', $('the remote-virtual-network-id of virtual network peering'))
+    .option('--allow-forwarded-traffic <allow-forwarded-traffic>', $('the allow-forwarded-traffic of virtual network peering'))
+    .option('--allow-gateway-transit <allow-gateway-transit>', $('the allow-gateway-transit of virtual network peering'))
+    .option('--use-remote-gateways <use-remote-gateways>', $('the use-remote-gateways of virtual network peering'))
+    .option('--allow-vnet-access <allow-vnet-access>', $('the allow-virtual-network-access of virtual network peering'))
+    .option('-s, --subscription <subscription>', $('the subscription identifier'))
+    .execute(function(resourceGroup, vnetName, name, options, _) {
+      if(!options.resourceGroup) {
```
amarzavery: resourceGroup, vnetName, name - All of these are positional as well as switch parameters.
Hence in ".execute(function(**resourceGroup, vnetName, name**, options, _) {" they are passed directly as parameters and not as options.

So you should be checking resourceGroup directly and not options.resourceGroup.

lib/commands/arm/network/virtualNetworkPeerings._js
```diff
@@ -0,0 +1,355 @@
...
+        parameters.virtualNetworkPeeringName = options.name;
+      }
+      generatorUtils.removeEmptyObjects(parameters);
+      progress = cli.interaction.progress(util.format($('Creating virtual network peering "%s"'), name));
+      try {
+        result = networkManagementClient.virtualNetworkPeerings.createOrUpdate(resourceGroup, vnetName, name, parameters, _);
+      } finally {
+        progress.end();
+      }
+      cli.interaction.formatOutput(result, generatorUtils.traverse);
```
amarzavery: does this take care about the json output when--json  switch is applied

amarzavery:
General Comment: The generated code is very well formatted and looks very much like hand written code. Keep up the good work!!

lib/commands/arm/network/virtualNetworkPeerings._js
MikhailTryakhov: Agree, Amar.
Currently options description are unusable. 
It's high priority task this week. Would plan to get descriptions from swagger

MikhailTryakhov:
Thanks @amarzavery 
good news that all your comments are fixable :) 
Would provide you updates next few days

lib/commands/arm/network/virtualNetworkPeerings._js
amarzavery: Awesome. Really good work! Reviewing this thoroughly  to make the generator better.

PR #3094, opened 2016-08-12
Change Output Location List

andrew-mestas:
* 
* 
*


azurecla:
Hi __@andrew-mestas__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (v-anmest). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/commands/arm/location/location._js
```diff
@@ -0,0 +1,106 @@
...
+
+var resourceLib = require('azure-arm-resource');
+
+var $ = utils.getLocaleString;
+
+exports.init = function (cli) {
+  var log = cli.output;
+  var withProgress = cli.interaction.withProgress.bind(cli.interaction);
+
+  var group = cli.category('location')
```
amarzavery: var location. 
correct naming helps in reading the code

lib/commands/arm/location/location._js
```diff
@@ -0,0 +1,106 @@
...
+    .description($('Commands to get the available locations'));
+  
+  group.command('list')
+    .description($('list the available locations'))
+    .option('--subscription <subscription>', $('the subscription identifier'))
+    .execute(withClient(function (client, options, _) {
```
amarzavery: why do you need the withClient function. This is straying away from the standard practice. I would like to reduce variations. New Service teams look at other commands as an example for implementing their commands. They copy over the style. Now we have different code paths to do the same thing which means more places for bugs to occur.

lib/commands/arm/location/location._js
```diff
@@ -0,0 +1,106 @@
...
+        function (log, _) {
+          providers = providerUtils.getAllProviders(client, _);
+        }, _);
+    if (!options.json) {
+      log.warn('The "location list" commands is changed to list subscription\'s locations. ' + 
+      'For old information, use "provider list or show" commands.');
+      log.info('Getting locations...');
+    }
+    var subscription = profile.current.getSubscription(options.subscription);
+    var subscriptionClient = new resourceLib.SubscriptionClient(subscription._createCredentials(),
```
amarzavery: Why would you want to do this over here. All of this is done in lib/utils.js. There are lot of methods over there that create different clients. Why do we need a new code path?

lib/commands/arm/location/location._js
```diff
@@ -0,0 +1,106 @@
...
+
+function showList(providerArr, numToShow){ 
+  var smallerList = providerArr.slice(0, numToShow).join(', ');
+  var ellipse =  providerArr.length > 4 ? '...' : '';
+    return smallerList + ellipse;
+}
+
+function getProvidersForLocation(providers, displayName) {
+  var providersByLocation = providers.filter(function(provider) {
+    if (provider.resourceTypes.filter(function(resourceType){
```
amarzavery: Can the code inside the if() conditional check be moved to a separate function to increase the code readability.

lib/commands/arm/location/location._js
andrew-mestas: I was using the other commands as a reference. In this case, we need both a resource and subscription client each get different lists. I'll work on refactoring and keep it within standards.

lib/commands/arm/location/location._js
andrew-mestas: There currently is no method within utils.js to create a subscription client. I'll work on adding this.

lib/commands/arm/location/location._js
andrew-mestas: Yes I'll work on that

andrew-mestas:
All comments addressed in 4bc76a4 and 3d21743.

lib/commands/arm/location/location._js
```diff
@@ -17,49 +17,52 @@
...
+  var location = cli.category('location')
     .description($('Commands to get the available locations'));
   
-  group.command('list')
+  location.command('list')
     .description($('list the available locations'))
     .option('--subscription <subscription>', $('the subscription identifier'))
-    .execute(function (options, callback) {
+    .execute(function (client, options, _) {
+    var providers;
```
TianoMS: indent.

lib/commands/arm/location/locationUtils.js
```diff
@@ -0,0 +1,56 @@
...
+             'displayName'  : location.displayName, 
+             'providers'    : getProvidersForLocation(providers, location.displayName) 
+            };
+  });
+};
+
+exports.showList = function (providerArr, numToShow){ 
+  var smallerList = providerArr.slice(0, numToShow).join(', ');
+  var ellipse =  providerArr.length > 4 ? '...' : '';
+    return smallerList + ellipse;
```
TianoMS: what if the total number of elements is less than 4? Do we still show "..." at the end?

lib/commands/arm/location/locationUtils.js
```diff
@@ -0,0 +1,56 @@
...
+
+function amountOfResourcesInLocation(provider, displayName) {
+  return provider.resourceTypes.filter(function(resourceType) { 
+      return ifLocationMatches(resourceType, displayName);
+    });
+}
+
+function ifLocationMatches(resourceType, displayName) {
+  return resourceType.locations.filter(function(locationName) { 
+        return locationName === displayName;
```
TianoMS: can we do it in a case-insensitive way?

lib/commands/arm/location/location._js
```diff
@@ -17,49 +17,52 @@
...
+    var subscription = profile.current.getSubscription(options.subscription);
+    var resourceClient = utils.createResourceClient(subscription);
+    var subscriptionClientOptions = {noSubscription: true};
```
TianoMS: do we need a variable for this?

TianoMS: Also, do we need to pass in noSubscription as true? I think it works with a subscription, too.

lib/commands/arm/location/locationUtils.js
andrew-mestas: It will just be the list without the '...'

lib/commands/arm/location/locationUtils.js
andrew-mestas: Yes I can add that

lib/commands/arm/location/location._js
andrew-mestas: To create a subscription client we need this variable to satisfy an if statement in the utils.js, this is one of the cases where a subscription is not needed when creating a REST client otherwise it tries to set a non-writeable parameter.

andrew-mestas:
Comments addressed in 66766b2. Travis CI build was failing previously from arm datalake commands.

TianoMS:
Looks good to me. @amarzavery can you take a look and merge this PR?

lib/commands/arm/location/location.js
```diff
@@ -1,65 +0,0 @@
...
-        return callback(err);
-      }
-      cli.interaction.formatOutput(result, function (data) {
-        if (data.length === 0) {
-          log.info($('No location found'));
-        } else {
-          log.table(data, function (row, location) {
-            row.cell($('Name'), location.name);
-            row.cell($('Display Name'), location.displayName);
-            row.cell($('Latitude'), location.latitude);
```
huangpf: Why removing the lovely latitude and longitude information?

andrew-mestas: They can be accessed now with the --details flag

PR #3099, opened 2016-08-16
use user-home lib in replace of HOME discover function

haocs:
* 
* 
*


amarzavery:
```
## Supported Environment Variables
Please take a look at the list of supported environment variables over [here](https://github.com/Azure/azure-xplat-cli/wiki/Environment-Variables).
```

amarzavery:
LGTM

PR #3100, opened 2016-08-16
help --json

amarzavery:
The content to be added to Changelog is as follows:
* support for help in json format

PR #3103, opened 2016-08-17
Improved the help command experience

amarzavery:
* Improved the help command experience


amarzavery:
#3100

Issue #2775, opened 2016-04-19
Azure CLI Test Infra doesn't support random values


huangpf:
CLI Version: 0.9.20
OS Type: Any
Installation via:  Any

Mode: Both

Description:
Azure CLI Test Infra doesn't support random values. It's blocking us from verifying the user caes of using completely automated/randomized resource names. Unlike PS & .NET SDK, they support it.

So I have to use hash function for this limitation:
* https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/vmss/vmss._js#L188

amarzavery:
@huangpf - Did you take a look at suite.generateId() method.
Several tests use it to create names for resources, resource groups, vms, etc. 
Take a look at some examples over here
- https://github.com/Azure/azure-xplat-cli/blob/dev/test/commands/arm/group/arm.group-tests.js#L68 
- https://github.com/Azure/azure-xplat-cli/blob/dev/test/commands/arm/role/arm.role-tests.js#L59 

We also provide a method `utils.uuidGen()` to generate Random GUID. This will generate a random GUID in record mode and will provide the same value in playback mode after restoring it from the recording file.

Take a look at some examples over here:
- https://github.com/Azure/azure-xplat-cli/blob/984313e0715f5b093955df532af5d48bf7828039/test/commands/arm/role/arm.role-tests.js#L219
- https://github.com/Azure/azure-xplat-cli/blob/984313e0715f5b093955df532af5d48bf7828039/test/commands/arm/role/arm.role-tests.js#L354

Hope this helps.

huangpf:
If I use ```utils.uuidGen()``` inside the command code, i.e. ```vm quick-create```, and then we write test code for this command, will it work in both record and playback modes? Just want to confirm. Thanks.

amarzavery:
The example above shows how it is used in the test. Whatever is used in the test is recorded. How can we record something that is used inside the command?

What is it that you are trying to achieve?

Is suite.generateId() useful for you?

huangpf:
Essentially, I'm looking for a way to test any generated resources inside the command itself, not the inputs to a command (which we can generate using the functions you gave).

I think the code, let say, ```vm._js``` would be like:
```
var oneNewRandomResource = generate();
```

And ```generate()``` would be:
```
function generate() {
  if (isPlayback() === true) {
    var result = readFromRecordQueue();
  }
  else if (isRecord() === true) {
    var result = generateImpl();
[...]

Issue #2963, opened 2016-06-27
azure vm quick-create: The storage account already taken


ahmetalpbalkan:
CLI Version: **0.10.0**
OS Type: Mac
Installation via:  npm

Mode: **ARM**

```
➜  ~ azure vm quick-create ahmet-rdma centos6 westus linux openlogic:CentOS-HPC:6.5:latest alp --ssh-publickey-file ~/.ssh/id_rsa.pub -z Standard_A8
info:    Executing command vm quick-create
+ Looking up the VM "centos6"
info:    Verifying the public key SSH file: /Users/alp/.ssh/id_rsa.pub
info:    Using the VM Size "Standard_A8"
info:    The [OS, Data] Disk or image configuration requires storage account
+ Looking up the storage account cli4265626714193486306
info:    Could not find the storage account "cli4265626714193486306", trying to create new one
+ Creating storage account "cli4265626714193486306" in "westus"
error:   The storage account named cli4265626714193486306 is already taken.
info:    Error information has been recorded to /Users/alp/.azure/azure.err
error:   vm quick-create command failed
```

I don't know what kind of randomization seed being used in coming up with that storage account name, but it is coming up the same name all the time (for the same command on different subs).

I was deleting a VM with the same name (and created using the same command) **in my another subscription** but storage account names overlapped while the deletion is still going on.

I suggest randomizing this a little bit more.

huangpf:
Currently, if the exactly same parameter input is used, it would generate the same storage name. One workaround is to change one of the inputs -- VM name, user name, RG name, image name, etc. -- even changing upper/lower cases may work...

ahmetalpbalkan:
> Currently, if the exactly same parameter input is used, it would generate the same storage name.

even on different subscriptions? that does not sound like a good idea.

huangpf:
The chance to get a conflict on storage account is more or less the same as someone stole one of your favorite random storage name, which should be minimal...

But with that said, we would definitely think about how to add more reasonable randomness in it, and yet deterministically testable...

huangpf:
@haocs @amarzavery @yugangw-msft 
Is there a way in the code to determine whether this is a test run or non-test run? In that way, in the test we can generate deterministic value, or completely random value in the other way.

ahmetalpbalkan:
@huangpf can't we just:

    hash(subscription+resourcegroup+vmname)

in this case if we get the same error it just means that we're very very unlucky.

huangpf:
[...]

PR #3104, opened 2016-08-17
random value generation that is production and test playback safe

amarzavery:
* ability to generate random values within a command; record them if a test runs that command and retrieve them from the recording file if the test for that command is being run in playback mode

### DIFF
--- a/lib/util/utils.js
+++ b/lib/util/utils.js
@@ -1250,6 +1250,16 @@ exports.uuidGen = function () {
   return uuid.v4();
 };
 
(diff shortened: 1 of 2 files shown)


amarzavery:
https://github.com/Azure/azure-xplat-cli/issues/2775
https://github.com/Azure/azure-xplat-cli/issues/2963

amarzavery:
@huangpf 
@ahmetalpbalkan 
This PR should fix #2775 and #2963.

huangpf:
Thanks @amarzavery!

PR #3112, opened 2016-08-20
Add vnet peerings implementation

EvgenyAgafonchikov:
- Updated azure-arm-network version
- Added vnet peerings implementation
- Added vnet peerings tests and corresponding recordings
- Fixed recordings broken after azure-arm-network version update


azurecla:
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@EvgenyAgafonchikov__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

MikhailTryakhov:
Hi @amarzavery, @yugangw-msft 
we've finally updated arm-network package and tests
please review generated vnet peerings

MikhailTryakhov:
Hi @DeepakRajendranMsft, @amarzavery, @yugangw-msft 
could you please merge it ASAP - it's important commit with updated recordings for azure-arm-network@0.16.0 instead of @0.13.2. Currently we're blocked to commit other functionality based on this package version

lib/commands/arm/network/virtualNetworkPeerings._js
```diff
@@ -0,0 +1,331 @@
...
+    .description($('Commands to manage virtual network peerings'));
+
+  virtualNetworkPeerings.command('create [resource-group] [vnet-name] [name]')
+    .description($('Create a virtual network peering'))
+    .usage('[options] <resource-group> <vnet-name> <name>')
+    .option('-g, --resource-group <resource-group>', $('the resource-group of virtual network peering'))
+    .option('-e, --vnet-name <vnet-name>', $('the virtual-network-name of virtual network peering'))
+    .option('-n, --name <name>', $('the virtual-network-peering-name of virtual network peering'))
+    .option('-r, --remote-vnet-id <remote-vnet-id>', $('the reference of the remote virtual network by Id'))
+    .option('-f, --allow-forwarded-traffic <allow-forwarded-traffic>', $('whether the forwarded traffic from the VMs in the remote virtual network will be allowed/disallowed'))
```
haocs: What's the expected value for this option? 'true'/'false' or 'allowed'/'disallowed'?

lib/commands/arm/network/virtualNetworkPeerings._js
```diff
@@ -0,0 +1,331 @@
...
+      }
+      if(!vnetName) {
+        vnetName = cli.interaction.promptIfNotGiven($('virtual-network-name : '), vnetName, _);
+      }
+      if(!name) {
+        name = cli.interaction.promptIfNotGiven($('virtual-network-peering-name : '), name, _);
+      }
+
+      if(!options.remoteVnetId) {
+        options.remoteVnetId = cli.interaction.promptIfNotGiven($('remote-virtual-network-id : '), options.remoteVnetId, _);
```
haocs: Is `remoteVnetId` a required argument? If yes, please specify it in `usage()`.

EvgenyAgafonchikov: Updated

lib/commands/arm/network/virtualNetworkPeerings._js
EvgenyAgafonchikov: Updated parameters with expected values

lib/commands/arm/network/virtualNetworkPeerings._js
```diff
@@ -0,0 +1,334 @@
...
+
+  var virtualNetworks = network.category('vnet')
+    .description($('Commands to manage virtual network peerings'));
+
+  var virtualNetworkPeerings = virtualNetworks.category('peering')
+    .description($('Commands to manage virtual network peerings'));
+
+  virtualNetworkPeerings.command('create [resource-group] [vnet-name] [name]')
+    .description($('Create a virtual network peering'))
+    .usage('[options] <resource-group> <vnet-name> <name> <--remote-vnet-id remote-vnet-id>')
```
haocs: `<--remote-vnet-id remote-vnet-id>`

lib/commands/arm/network/virtualNetworkPeerings._js
```diff
@@ -0,0 +1,334 @@
...
+    .option('-g, --resource-group <resource-group>', $('the resource-group of virtual network peering'))
+    .option('-e, --vnet-name <vnet-name>', $('the virtual-network-name of virtual network peering'))
+    .option('-n, --name <name>', $('the virtual-network-peering-name of virtual network peering'))
+    .option('-r, --remote-vnet-id <remote-vnet-id>', $('the reference of the remote virtual network by Id'))
+    .option('-f, --allow-forwarded-traffic <allow-forwarded-traffic>', $('whether the forwarded traffic from the VMs in the remote virtual network will be allowed/disallowed' +
```
haocs: On second thought, if expected option value is true or false, it is better to use a switch option(parameter). For example, if `--allow-forwarded-traffic` is provided, it will allow forwarded traffic. If the options is omitted, no forwarded traffic is allowed.

lib/commands/arm/network/virtualNetworkPeerings._js
```diff
@@ -0,0 +1,334 @@
...
+        name = cli.interaction.promptIfNotGiven($('virtual-network-peering-name : '), name, _);
+      }
+
+      var subscription = profile.current.getSubscription(options.subscription);
+      var networkManagementClient = utils.createNetworkManagementClient(subscription);
+      var result;
+
+      var progress = cli.interaction.progress(util.format($('Looking up the virtual network peering "%s"'), name));
+      try {
+        result = networkManagementClient.virtualNetworkPeerings.get(resourceGroup, vnetName, name, null, _);
```
haocs: Why do you need a get operation before delete?

EvgenyAgafonchikov: Delete operation doesn't throw exception in case user attempts to delete item that doesn't exist. This way we make operations transparent for user: they get error message while trying to delete non-existing item instead of silently skipping of real delete operation and displaying of OK message

MikhailTryakhov: Agree. That's a common style in CLI commands - throw exceptions before delete operation if item doesn't exist

lib/commands/arm/network/virtualNetworkPeerings._js
MikhailTryakhov: Good idea,
but common style in CLI works a bit different. We use true/false values here. Default value is false - that's why it work like a flag.
if any script would use such command - better to get user a chance to set value explicitly to make sure that functionality is expected

lib/commands/arm/network/virtualNetworkPeerings._js
EvgenyAgafonchikov: Could you please provide more details for this comment?

Currently I can suppose that it suggests to replace <--remote-vnet-id remote-vnet-id> with <remote-vnet-id>. In this case usage string could be confusing for users as the parameter looks like positional and no key should be specified before it.
Also, common approach in CLI implementation is to specify only positional required parameters in usage string (e.g. vnet subnet requires address-prefix however it is not specified in usage string).
Taking into account the above I suppose it would be better to return to the previous version of the usage string (w/o subnet-id) as it would suite common CLI style and shouldn't confuse users who are familiar with the current CLI version

haocs: `<--remote-vnet-id remote-vnet-id>` should be `<remote-vnet-id>`. It's not necessary to add -- for positional arguments.
For CLI, The common approach is to specify every required argument in usage message. Then for convenient purpose, you can provide an option for this argument(like `<name>` and `<resource-group>`). Otherwise, how could our users tell if an argument is required or not?

lib/commands/arm/network/virtualNetworkPeerings._js
haocs: I am afraid that I cannot agree on it. If you take a look at commands like `rm` and `git`, they are working in the same way as I suggested. For example if you say `rm -rf ` that means delete recursive and force. You don't have to say `rm -r "true" -f "true". The former is more explicit and shorter. It's quite common to use switch parameter when expected values are true or false.

lib/commands/arm/network/virtualNetworkPeerings._js
EvgenyAgafonchikov: Let's have a look in a counterexamples:
1. [Here](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L155) we prompt user to enter address-prefix. [There is](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L139) corresponding usage string
2. Similar for NSG Rule: [prompt](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L1339), [usage](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L1321)
3. Traffic manager: [prompt](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L1771), [usage](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L1753)
4. route-table route: [prompts](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L2076), [usage](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L2062)

the same for local-gateway, vpn-connection, application-gateway and some of its sub-items; express-route circuit, peering, authorization.

So I think absence of required parameters usage excluding name, parent name (if present), location. It looks like that in most cases this set is "\<Parameters of CreateOrUpdate function from corresponding dll\> + \<location\>"

lib/commands/arm/network/virtualNetworkPeerings._js
DeepakRajendranMsft: C:\windows\system32>azure network lb inbound-nat-rule create -h
help:    Add a load balancing inbound NAT rule to the load balancer
help:
help:    Usage: network lb inbound-nat-rule create [options] <resource-group> <lb-name> <name>
help:
help:    Options:
help:      -h, --help                                     output usage information
help:      -v, --verbose                                  use verbose output
help:      -vv                                            more verbose with debug output
help:      --json                                         use json output
help:      -g, --resource-group <resource-group>          the name of the resource group
help:      -l, --lb-name <lb-name>                        the name of the load balancer
help:      -n, --name <name>                              the name of the inbound NAT rule
help:      -p, --protocol <protocol>                      the rule protocol [tcp,udp]
help:      -f, --frontend-port <frontend-port>            the frontend port [0-65535]
help:      -b, --backend-port <backend-port>              the backend port [0-65535]
help:      -e, --enable-floating-ip <enable-floating-ip>  enable floating point ip [true, false]
help:      -i, --idle-timeout <idle-timeout>              the idle timeout specified in minutes
help:      -t, --frontend-ip-name <frontend-ip-name>      the name of the frontend ip configuration
help:      -s, --subscription <subscription>              the subscription identifier
help:
help:    Current Mode: arm (Azure Resource Management)

@haocs

in azure CLI we have always used true/false values here. we shouldnt be breaking how commands work. Perhaps we should consider changing all commands to your suggestion at a later point. For Networking commands, we strive to make sure that all commands are parameters are consistent so customers can have uniform experience

EvgenyAgafonchikov: In current CLI implementation there is another approach to bool flags:
[lb rule](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L571), [lb inbound-nat-rule](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L661), [nic](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/network/network._js#L936) and possible some other commands implemented this way. To keep code consistency after applying this approach in vnet peering we should rewrite every of the mentioned commands to use flags w/o parameter. However if CLI users are using these flags in their scripts, this would break their scripts.

lib/commands/arm/network/virtualNetworkPeerings._js
MikhailTryakhov: but current implemetation is incorrect. Please fix ASAP

MikhailTryakhov:
Hi @amarzavery, @haocs
could you please help us to finalize such discussion.
@DeepakRajendranMsft marked is as LGTM

This PR contains recordings which need to be merged **urgently**
It blocks me from other finctionality pushing.
Please help me ASAP

amarzavery:
@MikhailTryakhov - Please resolve merge  conflicts for the PR to be merged.

MikhailTryakhov:
@amarzavery, yeah, another comment have broken few our test recordings.
Currently it's fixed and ready to merge

PR #3113, opened 2016-08-22
disable validate authority

amarzavery:
* disable validate authority for AzureStack


amarzavery:
@shriramnat - This PR adds support for disabling validation authority for azure stack

PR #3123, opened 2016-08-28
Key Vault: Add support for certificates, move to auto-rest based node SK

rohmano:
1. Move to autorest base node SDK
2. Add support for Key Vault certificates
3. Key Vault certificates tests

### DIFF
--- a/lib/commands/arm/keyvault/keyvault._js
+++ b/lib/commands/arm/keyvault/keyvault._js
@@ -32,6 +32,7 @@ var RESOURCE_TYPE = 'Microsoft.KeyVault/vaults';
 var SKU_TYPE = ['Standard', 'Premium'];
 var KEY_PERMS = ['all', 'create', 'import', 'update', 'delete', 'get', 'list', 'backup', 'restore', 'sign', 'verify', 'encrypt', 'decrypt', 'wrapKey', 'unwrapKey'];
 var SECRET_PERMS = ['all', 'set', 'get', 'list', 'delete'];
+var CERTIFICATE_PERMS = ['all', 'get', 'list', 'delete', 'create', 'import', 'update', 'managecontacts', 'getissuers', 'listissuers', 'setissuers', 'deleteissuers'];
 
 exports.init = function(cli) {
 
@@ -220,7 +221,9 @@ exports.init = function(cli) {
               // Typical permissions for 'key manager' role - cannot sign, decrypt, etc. This default is same of Powershell.
               keys: ['get', 'create', 'delete', 'list', 'update', 'import', 'backup', 'restore'],
               // Typical permissions for 'secret manager' role. This default is same of Powershell.
-              secrets: ['all']
+              secrets: ['all'],
+              // Typical permissions for 'certificate manager' role. This default is same of Powershell.
+              certificates: ['all']
             }
           }];
 
@@ -488,6 +491,7 @@ exports.init = function(cli) {
     .option('--upn <user-principal-name>', $('name of a user principal that will receive permissions'))
     .option('--perms-to-keys <perms-to-keys>', util.format($('JSON-encoded array of strings representing key operations; each string can be one of [%s]'), KEY_PERMS.join(', ')))
     .option('--perms-to-secrets <perms-to-secrets>', util.format($('JSON-encoded array of strings representing secret operations; each string can be one of [%s]'), SECRET_PERMS.join(', ')))
+    .option('--perms-to-certificates <perms-to-certificates>', util.format($('JSON-encoded array of strings representing certificate operations; each string can be one of [%s]'), CERTIFICATE_PERMS.join(', ')))
     .option('--enabled-for-deployment <boolean>', $('specifies whether the Azure Compute resource provider can access secrets'))
     .option('--enabled-for-template-deployment <boolean>', $('specifies whether Azure Resource Manager can access secrets'))
     .option('--enabled-for-disk-encryption <boolean>', $('specifies whether Azure Disk Encryption can access secrets'))
@@ -511,21 +515,23 @@ exports.init = function(cli) {
 
       options.permsToKeys = kvUtils.parseArrayArgument('perms-to-keys', options.permsToKeys, KEY_PERMS, null);
       options.permsToSecrets = kvUtils.parseArrayArgument('perms-to-secrets', options.permsToSecrets, SECRET_PERMS, null);
+      options.permsToCertificates = kvUtils.parseArrayArgument('perms-to-certificates', options.permsToCertificates, CERTIFICATE_PERMS, null);
       options.enabledForDeployment = kvUtils.parseBooleanArgument('enabled-for-deployment', options.enabledForDeployment);
       options.enabledForTemplateDeployment = kvUtils.parseBooleanArgument('enabled-for-template-deployment', options.enabledForTemplateDeployment);
       options.enabledForDiskEncryption = kvUtils.parseBooleanArgument('enabled-for-disk-encryption', options.enabledForDiskEncryption);
 
-      if (!options.permsToKeys && !options.permsToSecrets && __.isUndefined(options.enabledForDeployment) && __.isUndefined(options.enabledForTemplateDeployment) && __.isUndefined(options.enabledForDiskEncryption)) {
+      if (!options.permsToKeys && !options.permsToSecrets && !options.permsToCertificates && __.isUndefined(options.enabledForDeployment) && __.isUndefined(options.enabledForTemplateDeployment) && __.isUndefined(options.enabledForDiskEncryption)) {
         log.error($('Please inform at least one of the following:'));
         log.error($('    --perms-to-keys <perms-to-keys>'));
         log.error($('    --perms-to-secrets <perms-to-secrets>'));
+        log.error($('    --perms-to-certificates <perms-to-certificates>'));
         log.error($('    --enabled-for-deployment <boolean>'));
         log.error($('    --enabled-for-template-deployment <boolean>'));
         log.error($('    --enabled-for-disk-encryption <boolean>'));
         throw new Error($('Inconsistent arguments'));
       }
 
-      if (options.permsToKeys || options.permsToSecrets) {
+      if (options.permsToKeys || options.permsToSecrets || options.permsToCertificates) {
 
         var v = [];
         if (options.objectId) v.push('--object-id');
@@ -556,7 +562,7 @@ exports.init = function(cli) {
         if (options.upn) extraneous = '--upn';
 
         if (extraneous) {
-          throw new Error(util.format($('Inconsistent arguments: %s was informed while none of [--perms-to-keys, --perms-to-secrets] was informed'), extraneous));
+          throw new Error(util.format($('Inconsistent arguments: %s was informed while none of [--perms-to-keys, --perms-to-secrets, --perms-to-certificates] was informed'), extraneous));
         }
 
       }
@@ -590,7 +596,7 @@ exports.init = function(cli) {
 
       var properties = vaultResource.properties;
 
-      if (options.permsToKeys || options.permsToSecrets) {
+      if (options.permsToKeys || options.permsToSecrets || options.permsToCertificates) {
 
         /////////////////////////////////////////////////////////////
         // Finds the policy to change, or add if not found.        //
@@ -608,7 +614,8 @@ exports.init = function(cli) {
             objectId: options.objectId,
             permissions: {
               keys: [],
-              secrets: []
+              secrets: [],
+              certificates: []
             }
           };
           policies.push(policy);
@@ -626,8 +633,12 @@ exports.init = function(cli) {
           policy.permissions.secrets = options.permsToSecrets;
         }
 
-        // If both are empty, the policy entry is not required.
-        if (!policy.permissions.keys.length && !policy.permissions.secrets.length) {
+        if (options.permsToCertificates) {
+          policy.permissions.certificates = options.permsToCertificates;
+        }
+
+        // If all are empty, the policy entry is not required.
+        if (!policy.permissions.keys.length && !policy.permissions.secrets.length && !policy.permissions.certificates.length) {
           var index = policies.indexOf(policy);
           policies.splice(index, 1);
         }
--- a/lib/commands/arm/keyvault/keyvault.certificate._js
+++ b/lib/commands/arm/keyvault/keyvault.certificate._js
@@ -1,0 +1,1753 @@
+/**
+ * Copyright (c) Microsoft.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+// Camel case checks disabled because the identifier certificate_ops is part of spec and can't be changed.
+/*jshint camelcase:false */
+
+var util = require('util');
+var fs = require('fs');
+
+var profile = require('../../../util/profile');
+var utils = require('../../../util/utils');
+var kvUtils = require('./kv-utils');
+var jsonlint = require('jsonlint');
+
+var $ = utils.getLocaleString;
+
+var ENCODINGS = ['base64'];
+var KEY_TYPES = ['RSA', 'RSA-HSM'];
+var SECRET_CONTENT_TYPES = ['application/x-pkcs12', 'application/x-pem-file'];
+
+exports.init = function(cli) {
+  var log = cli.output;
+
+  var certificate = cli.category('keyvault').category('certificate')
+    .description($('Commands to manage certificates in the Azure Key Vault service'));
+
+  certificate.command('list [vault-name]')
+    .description($('Lists certificates of a vault'))
+    .usage('[options] <vault-name>')
+    .option('-u, --vault-name <vault-name>', $('the vault name'))
+    .execute(function(vaultName, options, _) {
+
+      ///////////////////////
+      // Parse arguments.  //
+      ///////////////////////
+
+      log.verbose('arguments: ' + JSON.stringify({
+        vaultName: vaultName,
+        options: options
+      }));
+
+      options.vaultName = options.vaultName || vaultName;
+
+      if (!options.vaultName) {
+        return cli.missingArgument('vault-name');
+      }
+
+      ////////////////////////////////////////////
+      // Create the client and list certificates. //
+      ////////////////////////////////////////////
+
+      options.vaultUri = createVaultUri(options);
+      var client = createClient(options);
+      var certificates = [];
+      var progress = cli.interaction.progress(util.format($('Loading certificates of vault %s'), options.vaultUri));
+      try {
+        var result = client.getCertificates(options.vaultUri, _);
+        if (result) {
+          certificates = result;
+
+          while (result && result.nextLink) {
+            log.verbose(util.format($('Found %d certificates, loading more'), certificates.length));
+            result = client.getCertificatesNext(result.nextLink, _);
+            if (result) {
+              certificates = certificates.concat(result);
+            }
+          }
+        }
+      } finally {
+        progress.end();
+      }
+
+      log.table(certificates, showCertificateRow);
+
+      log.info(util.format($('Found %d certificates'), certificates.length));
+    });
+
+  certificate.command('list-versions [vault-name] [certificate-name]')
+    .description($('Lists certificate versions'))
+    .usage('[options] <vault-name> [certificate-name]')
+    .option('-u, --vault-name <vault-name>', $('the vault name'))
+    .option('-c, --certificate-name <certificate-name>', $('lists only versions of this certificate'))
+    .execute(function(vaultName, certificateName, options, _) {
+
+      ///////////////////////
+      // Parse arguments.  //
+      ///////////////////////
+
+      log.verbose('arguments: ' + JSON.stringify({
+        vaultName: vaultName,
+        certificateName: certificateName,
+        options: options
+      }));
+
+      options.vaultName = options.vaultName || vaultName;
+      options.certificateName = options.certificateName || certificateName;
+
+      if (!options.vaultName) {
+        return cli.missingArgument('vault-name');
+      }
+
+      ////////////////////////////////////////////
+      // Create the client and list certificates.       //
+      ////////////////////////////////////////////
+
+      options.vaultUri = createVaultUri(options);
+      var client = createClient(options);
+
+      var certificates, certificateIdentifier, certificateVersions;
+      var progress;
+      if (!options.certificateName) {
+        certificates = [];
+        progress = cli.interaction.progress(util.format($('Loading certificates of vault %s'), options.vaultUri));
+        try {
+          var result = client.getCertificates(options.vaultUri, _);
+          if (result && result.length) {
+            for (var i = 0; i < result.length; ++i) {
+              certificateIdentifier = kvUtils.parseCertificateIdentifier(result[i].id);
+              certificateVersions = getCertificateVersions(client, certificateIdentifier.vaultUri, certificateIdentifier.name, _);
+              certificates = certificates.concat(certificateVersions);
+            }
+
+            while (result && result.nextLink) {
+              log.verbose(util.format($('Found %d certificates, loading more'), certificates.length));
+              result = client.getCertificatesNext(result.nextLink, _);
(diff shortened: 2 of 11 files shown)


azurecla:
Hi __@rohmano__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (rohmano). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

rohmano:
@kunalcha, @pomortaz Please review

lib/commands/arm/keyvault/keyvault._js
```diff
@@ -32,6 +32,7 @@ var RESOURCE_TYPE = 'Microsoft.KeyVault/vaults';
 var SKU_TYPE = ['Standard', 'Premium'];
 var KEY_PERMS = ['all', 'create', 'import', 'update', 'delete', 'get', 'list', 'backup', 'restore', 'sign', 'verify', 'encrypt', 'decrypt', 'wrapKey', 'unwrapKey'];
 var SECRET_PERMS = ['all', 'set', 'get', 'list', 'delete'];
+var CERTIFICATE_PERMS = ['all', 'set', 'get', 'list', 'delete', 'manageissuers'];
```
pomortaz: Here is the list of permissions for certificates: "all", "get", "list", "delete", "create", "import", "update", "managecontacts", "getissuers", "listissuers", "setissuers", "deleteissuers", "manageissuers"

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+        return cli.missingArgument('vault-name');
+      }
+
+      ////////////////////////////////////////////
+      // Create the client and list certificates. //
+      ////////////////////////////////////////////
+
+      var client = createClient(options);
+      var certificates = [];
+      var progress = cli.interaction.progress(util.format($('Loading certificates of vault %s'), options.vaultUri));
```
amarzavery: I see options.vaultName being defined on line 57 but not vaultUri. When is vaultUri defined in the options object?

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+      try {
+        var result = client.getCertificates(options.vaultUri, null, _);
+        for (;;) {
+          if (!result) {
+              break;
```
amarzavery: indent 2 spaces

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+      ////////////////////////////////////////////
+      // Create the client and list certificates. //
+      ////////////////////////////////////////////
+
+      var client = createClient(options);
+      var certificates = [];
+      var progress = cli.interaction.progress(util.format($('Loading certificates of vault %s'), options.vaultUri));
+      try {
+        var result = client.getCertificates(options.vaultUri, null, _);
+        for (;;) {
```
amarzavery: Why do you need to have an infinite for loop. As you are using underscore as the last parameter (in place of the callback); streamline will replace the following code using `result` object correctly in the callback

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+          if (!result) {
+              break;
+            }
+
+          if (result.length) {
+            for (var i = 0; i < result.length; ++i) {
+              certificates = certificates.concat(result[i]);
```
amarzavery: concat joins two arrays and give the new array as a result. You don't need to use a for loop. Check this [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat)

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+        else {
+        currentCertificateContacts.contactList = [{ emailAddress: options.emailAddress}];
+        }
+
+        certificateContacts = client.setCertificateContacts(options.vaultUri, currentCertificateContacts, _);
+      } finally {
+        progress.end();
+      }
+
+      showCerti
```
amarzavery: Other person has no clue, that vaultUri is being set in the options object as a side effect of creating the client. It is a bad practice to modify the input and return something else. 
Please have a separate function named `constructVaultUri()` that does the job as specified by the function name.

lib/commands/arm/keyvault/keyvault.certificate._js
amarzavery: I see you are looping for getting the next page. The terminating condition for the loop could be a better one. 
```js
try {
  var result = client.getCertificates(options.vaultName,  _); 
  var certificates = result;
  while (result.nextLink) {
    result = client.getCertificatesNext(result.nextLink, _);
    certificates = certificates.concat(result);
  }
} finally {
  progress.end();
}
The above code pattern is more clear about the terminating condition. Please make sure to follow the above pattern in all the page-able list commands.

File: lib/commands/arm/keyvault/keyvault.certificate._js [hunk 7]
```diff
@@ -0,0 +1,1784 @@
...
+
+exports.init = function(cli) {
+  var log = cli.output;
+
+  var certificate = cli.category('keyvault').category('certificate')
+    .description($('Commands to manage certificates in the Azure Key Vault service'));
+
+  certificate.command('list [vault-name]')
+    .description($('Lists certificates of a vault'))
+    .usage('[--vault-name] <vault-name> [options]')
```
amarzavery: .usage pattern is incorrect. I understand that this was not noticed while reviewing the previous commands. However, let's correct it this time. Please take a look at the resource commands or group commands as an example and rectify the usages of all the keyvault commands. Basically, the [options] signifies the switch parameters. In the usage we only specify the required positional parameters explicitly.
In this case, the usage would have been: `.usage('[options] <vault-name>')`

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+        certificate = client.getCertificate(certificateIdentifier, _);
+      } finally {
+        progress.end();
+      }
+
+      options.passThru = true;
+      showCertificate(options, certificate);
+    });
+
+  certificate.command('get [vault-name] [certificate-name] [certificate-version] [file]')
```
amarzavery: Please name the command `download`. That makes it more clear than `get`.

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+      var progress = cli.interaction.progress(util.format($('Getting certificate %s'), certificateIdentifier));
+      try {
+        certificate = client.getCertificate(certificateIdentifier, _);
+      } finally {
+        progress.end();
+      }
+      
+      log.info(util.format($('Writing certificate to %s'), options.file));
+      var data = certificate.cer;
+      if (options.encode == 'base64') {
```
amarzavery: javascript uses `===`

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+
+    });
+
+  certificate.command('create [vault-name] [certificate-name]')
+    .description($('Creates a certificate in the vault'))
+    .usage('[--vault-name] <vault-name> [--certificate-name] <certificate-name> [options]')
+    .option('-u, --vault-name <vault-name>', $('the vault name'))
+    .option('-c, --certificate-name <certificate-name>', $('the certificate name'))
+    .option('-p --certificate-policy <certificate-policy>', $('a JSON-formatted string containing the certificate policy'))
+    .fileRelatedOption('-e --certificate-policy-file <certificate-policy-file>', $('a file containing the certificate policy'))
```
amarzavery: It's worthwhile mentioning in the help that this option is mutually exclusive with --certificate-policy

lib/commands/arm/keyvault/keyvault.certificate._js
amarzavery: Actually, since there is an import command, it will make more sense to name this command as `export`

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+
+  certificate.command('import [vault-name] [certificate-name]')
+    .description($('Imports a certificate into the vault'))
+    .usage('[--vault-name] <vault-name> [--certificate-name] <certificate-name> [options]')
+    .option('-u, --vault-name <vault-name>', $('the vault name'))
+    .option('-c, --certificate-name <certificate-name>', $('the certificate name'))
+    .option('-b --content <content>', $('a base64 encoded string containing the certificate and private key (optional) to be imported'))
+    .fileRelatedOption('-e --content-file <content-file>', $('a file containing the certificate and private key (optional) to be imported'))
+    .option('--content-file-binary <boolean>', $('indicates if the content file is binary. valid values: [false, true]. default is false.'))
+    .option('-s, --password <password>', $('password of the file if it contains a encrypted private key'))
```
amarzavery: -s is reserved for subscription and cannot be used for anything else

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+      showCertificate(options, certificate);
+    });
+
+  certificate.command('set-attributes [vault-name] [certificate-name] [certificate-version]')
+    .description($('Changes attributes of an existing certificate'))
+    .usage('[--vault-name] <vault-name> [--certificate-name] <certificate-name> [[--certificate-version] <certificate-version>] [options]')
+    .option('-u, --vault-name <vault-name>', $('the vault name'))
+    .option('-c, --certificate-name <certificate-name>', $('name of the certificate to be modified'))
+    .option('-r, --certificate-version <certificate-version>', $('the version to be modified; if omited, modifies only the most recent'))
+    .option('--enabled <boolean>', $('if informed, command will change the enabled state; valid values: [false, true]'))
```
amarzavery: you can have a boolean switch. It need not accept any value. If the user wants to enable something then he/she can simply provide `--enable` flag. Please take a look at the telemetry command switches to get an idea.

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+      if (options.file) {
+        fs.writeFileSync(options.file, JSON.stringify(policy), { flag: 'wx' });
+      }
+      else {
+        options.passThru = true;
+        showCertificatePolicy(options, policy);
+      }
+    });
+
+  certificatePolicy.command('update [vault-name] [certificate-name]')
```
amarzavery: please use set as the command name. That ensures consistency. The documentation folder of this repo has lot of useful information that talks about allowed verbs and recommended patterns for designing commands.

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+        certificatePolicy = client.getCertificatePolicy(options.vaultUri, options.certificateName, _);
+      } finally {
+        progress.end();
+      }
+
+      options.passThru = true;
+      showCertificatePolicy(options, certificatePolicy);
+    });
+
+  certificatePolicy.command('get [vault-name] [certificate-name] [file]')
```
amarzavery: please name the command show

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+        certificate = client.deleteCertificate(options.vaultUri, options.certificateName, _);
+      } finally {
+        progress.end();
+      }
+
+      options.passThru = true;
+      showCertificate(options, certificate);
+    });
+
+  var certificatePolicy = cli.category('keyvault').category('certificate-policy')
```
amarzavery: policy should actually be a sub category inside certificate. This will make things more CLI like and less powershelly.

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+        certificatePolicy = client.updateCertificatePolicy(options.vaultUri, options.certificateName, certificatePolicy, _);
+      } finally {
+        progress.end();
+      }
+
+      options.passThru = true;
+      showCertificatePolicy(options, certificatePolicy);
+    });
+
+  var certificateOperation = cli.category('keyvault').category('certificate-operation')
```
amarzavery: operation should be a category inside certificate

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+        certificateOperation = client.deleteCertificateOperation(options.vaultUri, options.certificateName, _);
+      } finally {
+        progress.end();
+      }
+
+      options.passThru = true;
+      showCertificateOperation(options, certificateOperation);
+    });
+
+  var certificateAdministratorDetails = cli.category('keyvault').category('certificate-administrator-details')
```
amarzavery: 1. It should be a category inside certificate
2. The category name should simply be administrator
`azure keyvault certificate administrator create` looks better and consistent with the design of other commands than `azure keyvault certificate-administrator-details create`

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1784 @@
...
+      // Create certificate administrator details //
+      ////////////////////////////////////////////
+      var administratorDetails = {
+        firstName : options.firstName,
+        lastName : options.lastName,
+        emailAddress : options.emailAddress,
+        phone : options.phoneNumber
+      };
+
+      if (options.file) {
```
amarzavery: what kind of command is this? It never calls the keyvault client? What is the command trying to achieve?

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: method createClient defines options.vaultName

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Semantically, export is incorrect for this command, especially since it deals with a certificate.
I am trying to be consistent across key vault, where get has been used to download.
I would like to maintain that consistency. Do you feels this warrants a breaking change?

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Trying to maintain consistency across key vault (see secrets) where show displays the object on the cmdline and get writes to disk. I would like to maintain consistency.
Do you feel this warrants a breaking change?

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: In my mind set translates to HTTP PUT. This operation underneath does a HTTP PATCH so update seems more appropriate.

In other words, I would think that when doing a set, all of the input params would be needed as compared to a specific one(s) when doing update.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: The command is helping the user to work with complex objects. The actual use of this command is to create a certificate-issuer object on the server. The certificate issuer is a complex object and it can be tricky for the users to construct one. This command intends to mimic some of the piping behavior available in PS. Use this command to write JSON to disk. Feed that JSON to another command...

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault._js
rohmano: Fixed. Omitted "manageissuers" since this is only for legacy reasons. We don't want new users to use this.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Even for non ARM operations?

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: The user should be able to disable a certificate too.

lib/commands/arm/keyvault/keyvault.certificate._js
amarzavery: I was confused between a get and a show. To me download is more clear. It signifies that outcome of this command will be some sort of file on my machine.

lib/commands/arm/keyvault/keyvault.certificate._js
amarzavery: I know that this is data plane. However, it is better to not use the reserved switches for anything else. In that way it does not confuse the user when he uses other CLI commands (which are not data plane).

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1783 @@
...
+      }
+
+      fs.writeFileSync(options.file, JSON.stringify(certificatePolicy), { flag: 'wx' });
+    });
+
+  certificatePolicy.command('create [issuer-name] [subject-name] [file]')
+    .description($('Creates a new certificate policy'))
+    .usage('[--issuer-name] <issuer-name> [[--subject-name] <subject-name>] [[--file] <--file>] [options]')
+    .option('-i, --issuer-name <issuer-name>', $('issuer name'))
+    .option('-s, --subject-name <subject-name>', $('subject name'))
```
amarzavery: -s reserved for subscription

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1783 @@
...
+        },
+        x509CertificateProperties: {
+        },
+        issuerReference: {
+          name: options.issuerName
+        }
+      };
+
+      if (options.keyType) {
+        if (!KEY_TYPES.some(function(e) { return e.toUpperCase() == options.keyType.toUpperCase(); })) {
```
amarzavery: === in javascript land http://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1783 @@
...
+          throw new Error(util.format($('Invalid value for key type argument. Accepted values are: %s'), KEY_TYPES.join(', ')));
+        }
+
+        policy.keyProperties.keyType = options.keyType.toUpperCase();
+      }
+
+      if (options.secretContentType) {
+        if (!SECRET_CONTENT_TYPES.some(function(e) { return e == options.secretContentType; })) {
```
amarzavery: same as above

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Resolved offline that we will maintain current usage to maintain consistency and avoid breaking change.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Resolved offline that we will maintain current usage to maintain consistency and avoid breaking change.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed. Updated help to indicate usage in a better way.

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1790 @@
...
+      options.vaultUri = createVaultUri(options);
+      var client = createClient(options);
+
+      var certificates, certificateIdentifier, certificateVersions;
+      var progress;
+      if (!options.certificateName) {
+        certificates = [];
+        progress = cli.interaction.progress(util.format($('Loading certificates of vault %s'), options.vaultUri));
+        try {
+          var result = client.getCertificates(options.vaultUri, null, _);
```
amarzavery: it is not required to pass null as the second last parameter for options. Its an optionsOrCallback kind of a things. The generated code will understand what you have passed.

lib/commands/arm/keyvault/keyvault.certificate._js
amarzavery: -s not yet removed.

rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
rohmano: Fixed.

lib/commands/arm/keyvault/keyvault.certificate._js
```diff
@@ -0,0 +1,1748 @@
...
+      try {
+        certificate = client.getCertificate(certificateIdentifier, _);
+      } finally {
+        progress.end();
+      }
+
+      showCertificate(options, certificate);
+    });
+
+  certificate.command('get [vault-name] [certificate-name] [certificate-version] [file]')
```
pomortaz: I think you should use 'show' instead of 'get' here.

amarzavery: I got confused initially. But this command downloads the certificate. There is a show command that displays the object returned from the service. 

If the KV team is ok with breaking changes then please rename all the commands to download. This makes things more clear.

amarzavery: We will try to add support for aliasing in the next release. At that time we can alias this command to download. In that way it would be less breaking for customers.

rohmano: @pomortaz Currently secrets use 'get' for download, so to maintain consistency certificates do the same. How do you feel about a breaking change? This is secret download, probably our most used method

pomortaz: @rohmano thanks for explaining. Based on what @amarzavery said we can perhaps later on introduce an alias for the command to support backward compatibility for secrets. I don't recommend introducing the breaking changes if we had adaptation with no issue reported. I will leave naming of certificate up to you. I think it makes sense to have consistency with secrets APIs.

PR #3125, opened 2016-08-30
Fix #3118: Add '-w, --public-ip-domain-name' to 'azure vm quick-create'

### LINKED ISSUE #3118: `azure vm quick-create` needs option for DNS label
- CLI Version: 0.10.3
- OS Type: Linux (RedHat)
- Installation via:  npm
- Mode: ARM
- Environment: AzureCloud

Description:
I'd like to see the `azure vm quick-create` command have an option for the DNS label, similar to the `-w, --public-ip-domain-name` option for the `azure vm create` command.

This would make automation much easier, and prevent me from having to use the `azure vm network public-ip set --domain-name-label` command after using quick-create.

huangpf:
* fixes #3118 
* 
*


huangpf:
@amarzavery @haocs Could you please help review? 

+ @sauryadas FYI

PR #3131, opened 2016-09-01
update location and recordings to fix iothub tests

EvgenyAgafonchikov:
This fixes iothub tests in Live\Recording mode: Previously they took location value from previous tests (process.env setup). The location taken is not supported for iothub operations.

### DIFF
--- a/test/commands/arm/iothub/arm.iothub-tests.js
+++ b/test/commands/arm/iothub/arm.iothub-tests.js
@@ -33,7 +33,7 @@ var knownNames = [];
 var requiredEnvironment = [{
   requiresToken: true
 }, {
-  name: 'AZURE_ARM_TEST_LOCATION',
+  name: 'AZURE_ARM_IOTHUB_TEST_LOCATION',
   defaultValue: 'West US'
 }, {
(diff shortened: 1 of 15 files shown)


azurecla:

TTYL, AZPRBOT;

test/commands/arm/iothub/arm.iothub-tests.js
```diff
@@ -33,7 +33,7 @@ var knownNames = [];
 var requiredEnvironment = [{
   requiresToken: true
 }, {
-  name: 'AZURE_ARM_TEST_LOCATION',
+  name: 'AZURE_ARM_IOTHUB_TEST_LOCATION',
```
amarzavery: If the location is "West US", then it is ok, to have one environment variable 'AZURE_ARM_TEST_LOCATION' rather than creating a big list of environment variables as it is hard to maintain them in the long run.

can you tell us why do we need to create separate service specific location environment variables.

EvgenyAgafonchikov: https://github.com/Azure/azure-xplat-cli/pull/3136#discussion_r78328743

PR #3136, opened 2016-09-05
Fix HDInsights test failures

EvgenyAgafonchikov:
* Changed location to supported
* Added creation of required storage account
* Initialize parameters with default values if env is not set


azurecla:

TTYL, AZPRBOT;

test/commands/arm/hdinsight/arm.hdinsight-cluster-crud-tests.js
```diff
@@ -78,10 +78,10 @@ describe('arm', function () {
     before(function (done) {
       suite = new CLITest(this, testprefix, requiredEnvironment);
       suite.setupSuite(function () {
-        location = process.env.AZURE_ARM_TEST_LOCATION;
-        defaultStorageAccount = process.env.AZURE_ARM_TEST_STORAGEACCOUNT;
-        defaultStorageAccountKey = process.env.AZURE_STORAGE_ACCESS_KEY;
-        defaultStorageContainer = process.env.AZURE_STORAGE_CONTAINER;
+        location = process.env.AZURE_ARM_HDI_TEST_LOCATION || location;
```
amarzavery: please make sure to add the new environment variables to the lsit of supported environemnt variables in the documentation folder.

test/commands/arm/hdinsight/arm.hdinsight-cluster-crud-tests.js
```diff
@@ -29,8 +29,8 @@ var appNamePrefix = "xplatSampleApp";
 var HdinsightTestUtil = require('../../../util/hdinsightTestUtil');
 var requiredEnvironment = [
   {
-    name: 'AZURE_ARM_TEST_LOCATION',
-    defaultValue: 'eastus'
+    name: 'AZURE_ARM_HDI_TEST_LOCATION',
```
amarzavery: Why a new environment variable is being added (AZURE_ARM_HDI_TEST_LOCATION) ?

EvgenyAgafonchikov: If we have no environment variable AZURE_ARM_TEST_LOCATION specified directly it is set in the first test suite and then used during full test run, next tests do not override it. However some of the tests are location dependent.
Also, it makes some tests depending on the order of running. Say, I'd like to split test list to run tests in parallel to save time. If I chose the first group before DataLake tests and the second starting from DataLake tests, in the second group I have `AZURE_ARM_TEST_LOCATION == "East US 2"` that is not supported for the next Iothub tests.
One way to resolve this is
- Adding of new environment variables for location dependent tests.

Another ways I can suggest are:
- Specifying of the location directly for location-dependent tests. In this case it wouldn't be overridden by environment variables at all.
- Creating of mapping of the supported locations (or checking them using `azure location list`, in this case we likely have to store related provider names in tests), adding of check in the beginning of each test and warning user that the location is not supported for this commands if they have unsupported value in the environment variable.

Please choose an option that you consider as the best one. If you have any other ideas on this, please let me know.

test/commands/arm/hdinsight/arm.hdinsight-cluster-crud-tests.js
EvgenyAgafonchikov: Let's decide what can we make on this
https://github.com/Azure/azure-xplat-cli/pull/3136#discussion_r78328743

PR #3152, opened 2016-09-12
arm webapp additions:

LukaszStem:
-edit: appserviceplan create help did not show all options (omitted standard S sizes)
-edit: webapp config show now displays more information
-edit: webapp config show can also display extra custom properties with --propertiesToInclude
-edit: webapp config set can now change multiple options in one command
-add: webapp config appsetting set command
-add: webapp config appsetting delete command
-add: webapp config appsetting list command

**Note:**
 In the last release,the --appsetting option was within the webapp config set command and did not work
 Now it has it's own separate command all together.
 Since previous functionality did not exist, I conclude that this is not a breaking change


lib/commands/arm/appserviceplan/appserviceplan._js
```diff
@@ -55,9 +56,9 @@ exports.init = function (cli) {
     .execute(function(resourceGroup, name, location, tier, options, _) {
       var subscription = profile.current.getSubscription(options.subscription);
       var client = appserviceplanUtils.createWebappManagementClient(subscription);
-	  var numOfInstances;
-	  var sku;
-	  var workerSize;
+      var numOfInstances;
```
yugangw-msft: One question, what power-shell does is to have the `tier` take values like `Basic, Free, etc`, and expose an extra parameter of `worker-size` with value of `Small, Medium`. I am fine with either ways as long as we know this. IMO, i think what we have is better as the "B1", "B2" are what portal displays right now.

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -30,11 +30,14 @@ exports.init = function (cli) {
...
 
+  var appsetting = config.category('appsetting')
+    .description($('Commands to manage your Azure webapps app setting configurations'));
+
   var publish = webapp.category('publishprofile')
     .description($('Command to get your Azure webapps publishing profile'));
 
   config.command('show [resource-group] [name]')
-    .description($('Get webapp configuration'))
+    .description($('Get webapp configuration \nexample:  webapp config show resroucegroup1 webapp1\nexample:  webapp config show resroucegroup1 webapp1 --propertiesToInclude remoteDebuggingVersion,logsDirectorySizeLimit '))
```
yugangw-msft: Can we offer a complete list extra property bag name for people to choose? It is hard for people to figure out the exact name, at least for me :-)

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -50,53 +53,98 @@ exports.init = function (cli) {
...
+          log.data($('Java version                   :'), data.javaVersion);
+          log.data($('Java Container                 :'), data.javaContainer);
+          log.data($('Java Container Version         :'), data.javaContainerVersion);
+        }
+      });
+      cli.interaction.formatOutput(publishingResult, function (data) {
+        if (!data) {
+          log.info($('No webapp publishing profile information available'));
+        } else {
+          log.data($('Publishing Username            :'), data.publishingUserName);
```
yugangw-msft: suggest name as `Publish Profile Credentials`, to distinguish from `user credentials`, per [kudu doc] (https://github.com/projectkudu/kudu/wiki/Deployment-credentials)

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -50,53 +53,98 @@ exports.init = function (cli) {
...
+        if(appSettingsResult.properties !== null && appSettingsResult.properties !== undefined) {
+           log.data($('App Settings (use webapp config appsetting to change)'));
+           log.data($('--------------------------------'));
+           log.data(data);
+           log.data($(''));
+
+          }
+      });
+      cli.interaction.formatOutput(result, function (data) {
+        if(options.propertiesToInclude !== null && options.propertiesToInclude !== undefined) {
```
yugangw-msft: Can this be simplified as `if  options.propertiesToInclude`?

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -50,53 +53,98 @@ exports.init = function (cli) {
...
 	  
   config.command('set [resource-group] [name]')
-    .description($('set webapp configuration (1 setting at a time)'))
-    .usage('[options] <resource-group> <name>')
+    .description($('set webapp configuration'))
+    .usage('[options] <resource-group> <name>\nexample:  webapp config set resroucegroup1 webapp1 --alwayson true --numberofworkers 1')
     .option('-g --resource-group <resource-group>', $('the name of the resource group'))
     .option('-n --name <name>', $('the name of the webapp to update'))
-    .option('-l --location <location>', $('the location of the webapp to update, , \n\t\t\t\t\t\t   locations found at https://azure.microsoft.com/en-us/status/'))
+    .option('-l --location <location>', $('the location of the webapp to update,\n\t\t\t\t\t\t\tlocations found at https://azure.microsoft.com/en-us/status/'))
```
yugangw-msft: Do we really need to expose the `location`? we can get it from the webapp, right?

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -50,53 +53,98 @@ exports.init = function (cli) {
...
+            log.data($(''));
+          }
+      });
     });
 	  
   config.command('set [resource-group] [name]')
-    .description($('set webapp configuration (1 setting at a time)'))
-    .usage('[options] <resource-group> <name>')
+    .description($('set webapp configuration'))
+    .usage('[options] <resource-group> <name>\nexample:  webapp config set resroucegroup1 webapp1 --alwayson true --numberofworkers 1')
```
yugangw-msft: Question, do we have grand-child properties (nested under a parent node)? If yes, how users can update it?

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -50,53 +53,98 @@ exports.init = function (cli) {
...
-    .option('--appcommandline <appcommandline>', $('app command line for webapp'))
-    .option('--javaversion <javaversion>', $('javaversion of webapp'))
-    .option('--javacontainer <javacontainer>', $('javacontainer of webapp'))
-    .option('--javacontainerversion <javacontainerversion>', $('javacontainerversion of webapp'))
-    .option('--appsettings <appsettings>', $('webapp appsettings'))
+    .option('--phpversion <phpversion>', $('php version of webapp'))
+    .option('--pythonversion <pythonversion>', $('python version of webapp'))
+    .option('--nodeversion <nodeversion>', $('node version of webapp'))
+    .option('--numberofworkers <numberofworkers>', $('node version of webapp'))
+    .option('--netframeworkversion <netframeworkversion>', $('node version of webapp'))
```
yugangw-msft: copy & paste error? Please update several ones below :-)

lib/commands/arm/appserviceplan/appserviceplan._js
LukaszStem: I've decided this to be a more elegant way with less parameters.

 (sample of appserviceplan create -h)

help:    Create an App Service Plan
help:
help:    Usage: appserviceplan create [options] <resource-group> <name> <location> <tier>
help:
help:    Options:
help:      .......
help:      -t --tier <tier>                      tier of the App Service Plan
help:           eg:
help:                    Free, Shared,
help:                    B1 (Basic Small),    B2 (Basic Medium),    B3 (Basic Large),
help:                    S1 (Standard Small), S2 (Standard Medium), S3 (Standard Large),
help:                    P1 (Premium Small),  P2 (Premium Medium),  P3 (Premium Large)
help:
help:      -i --instances <numberOfInstances>    the number of instances in the App Service Plan, Default is 1
help:     ....

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -135,33 +219,225 @@ exports.init = function (cli) {
...
+        try {
           result = client.sites.getSite(resourceGroup, name, _);
         } finally {
           progress.end();
         }
-	    jsonObj['location'] = result.location;
-	  }
+	      jsonObj['location'] = result.location;
+      }
+      if( !(Object.keys(jsonObj).length === 0 && jsonObj.constructor === Object)) {
```
yugangw-msft: Because the `jsonObj` starts with `{}`, i don't think the sub-condition of `jsonObj.constructor === Object` is needed

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -135,33 +219,225 @@ exports.init = function (cli) {
...
+      else
+      {
+        log.info('No options selected');
+        progress.end();
+      }
+    });
+
+  function toBool(stringToParse) {
+    var str = stringToParse.toUpperCase();
+    if(str == 'TRUE') return true;
```
yugangw-msft: you can use `util.ignoreCaseEquals` like other commands. Also should you throw on other values?

lib/commands/arm/appserviceplan/appserviceplan._js
yugangw-msft: could you also confirm with your team, whether you do want to surface `appserviceplan` to the top level, or wrap it under `webapp`. I will let python version to follow the decision

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -135,33 +219,225 @@ exports.init = function (cli) {
...
+    if(str == 'FALSE') return false;
+    return stringToParse;
+  }
+
+   appsetting.command('list [resource-group] [name]')
+    .description($('Get webapp app settings'))
+    .usage('[options] <resource-group> <name>')
+    .option('-g --resource-group <resource-group>', $('the name of the resource group'))
+    .option('-n --name <name>', $('the name of the webapp to show'))
+    .option('--slot <slot>', $('the name of the slot to show'))
```
yugangw-msft: The `slot` doesn't appear used, but i assume your change in future will check the value and if not none, you will invoke the invoke the `listSiteAppSettingsSlot`?

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -135,33 +219,225 @@ exports.init = function (cli) {
...
+      var subscription = profile.current.getSubscription(options.subscription);
+      var client = webappUtils.createWebappManagementClient(subscription);
+      var progress = cli.interaction.progress($('Getting webapp configuration'));
+      var currentAppSettings;
+      try {
+        currentAppSettings = client.sites.listSiteAppSettings(resourceGroup, name, _);
+      } finally {
+        progress.end();
+      }
+      cli.interaction.formatOutput(currentAppSettings.properties, function (data) {
```
yugangw-msft: since it is about `add`, i don't think we should display the existing ones

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -135,33 +219,225 @@ exports.init = function (cli) {
...
+        settingSplit = appSettingsToAddSplit[i].split('=');
+        appSettingsProperties[settingSplit[0]] = settingSplit[1];
+      }
+      actualAppSettings.properties = appSettingsProperties;
+      try {
+        result = client.sites.getSite(resourceGroup, name, _);
+      } finally {
+        progress.end();
+      }
+	    actualAppSettings.location = result.location;
```
yugangw-msft: indent

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -135,33 +219,225 @@ exports.init = function (cli) {
...
+        progress.end();
+      }
+      log.data($('Final App Settings'));
+      log.data($('--------------------------------'));
+      log.data(result.properties);
+      log.data($(''));
+    });
+
+  appsetting.command('delete [resource-group] [name] [appsettings]')
+    .description($('Delete webapp app settings (using comma seperated keys)\nexample:  webapp config appsetting add KEY1,KEY2,KEY3'))
```
yugangw-msft: Suggest call out the complete command line with resource group name and site name

yugangw-msft:
:clock1:

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -135,33 +219,225 @@ exports.init = function (cli) {
...
+          log.data(data);
+          log.data($(''));
+        }
+        else {
+          log.data($('No app setting set'));
+        }
+      });
+    });
+
+   appsetting.command('add [resource-group] [name] [appsettings]')
```
yugangw-msft: Do you plan to have a `set`? If yes, then it might make sense to have a `set` to handle both

lib/commands/arm/appserviceplan/appserviceplan._js
ruslany: appserviceplan should not be under webapp. It is a separate entity in ARM and in our API.

yugangw-msft: @ruslany, have you considered having the `appservice` as the root category, with `web`, `plan` as children? And in future you can add other services such as 'mobile', `logic` underneath? To surface both to the top might make it harder for people to logically associate them or even find them among 30+other services.

lib/commands/arm/webapp/webapp.config._js
LukaszStem: I spoke with Ruslan and and we decided to remove propertiesToInclude and instead just show everything.

lib/commands/arm/webapp/webapp.config._js
LukaszStem: Yes

lib/commands/arm/webapp/webapp.config._js
LukaszStem: the string will return and show as invalid anyway

lib/commands/arm/appserviceplan/appserviceplan._js
LukaszStem: Yugang makes a good point. That way it better enforces the mentality that app service plans and web apps are closely related. But technically yes it is a separate ARM entity. I'll leave this up to Ruslan as he has much greater knowledge.

lib/commands/arm/webapp/webapp.config._js
LukaszStem: I think we should.. I do the same for delete as well. In my opinion it better informs the user... what is your reasoning?

lib/commands/arm/webapp/webapp.config._js
LukaszStem: I moved java version information to be towards the bottom.

yugangw-msft: I meant the description is not correct, say ` .option('--numberofworkers <numberofworkers>', $('node version of webapp'))`

lib/commands/arm/webapp/webapp.config._js
yugangw-msft: The scenario will be a user uses `list` to see the value, and then a `set` to update it;  so they should  care whether the value is update or not. You can choose to display all values after the update, which is more meaningful than doing before. Also `list` is always there for people to cross check.

lib/commands/arm/webapp/webapp.config._js
LukaszStem: Nope, just add/delete/list

lib/commands/arm/appserviceplan/appserviceplan._js
yugangw-msft: @ruslany, if not further question/comments, let us just go with top category of `appservice`, with `web` and `plan` underneath, and `logic`, `mobile` in future. Lukasz doesn't need to make the change in this PR, but get it done by the September release

lib/commands/arm/webapp/webapp.config._js
yugangw-msft: Okay, as long as the service return good error, that would be fine

yugangw-msft: One question though, have you thought about be in more parity with portal with flags of `on` and `off`; rather `true` or `false`?

lib/commands/arm/webapp/webapp.config._js
yugangw-msft: But is it a valid scenario that user want to update the value? Do you expect them to use `add`?

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -135,33 +215,195 @@ exports.init = function (cli) {
...
+      var settingSplit;
+      var result;
+      var appSettingsProperties = {};
+      var actualAppSettings = {};
+      for (var attrname in appSettingsPrevious) { appSettingsProperties[attrname] = appSettingsPrevious[attrname]; }
+      appSettingsToAddSplit = appSettingsToAdd.toString().split(',');
+      settingSplit = [];
+      for(var i = 0; i < appSettingsToAddSplit.length; i++) {
+        settingSplit = appSettingsToAddSplit[i].split('=');
+        appSettingsProperties[settingSplit[0]] = settingSplit[1];
```
yugangw-msft: is it possible that the value could contain `=`? If yes, then ` settingSplit[1]` is not a complete value. I suggest you split at the first `=` and done.

yugangw-msft:
Added some more comments for clarification. The rest LGTM

lib/commands/arm/webapp/webapp.config._js
LukaszStem: You make a valid point. Will change to set.

lib/commands/arm/appserviceplan/appserviceplan._js
ruslany: Hi Yugang, we discussed this within a team and decided that we do not want to use top category appservice. We want the web and appserviceplan to remain as they are. That way they are easily discoverable.

yugangw-msft: Okay.

lib/commands/arm/webapp/webapp.config._js
LukaszStem: example if I accidently typed in falsse -> 
Error "siteConfig.properties.alwaysOn with value falsse must be of type boolean." occurred in serializing the payload - "{ alwaysOn: 'falsse', location: 'South Central US' }"
I will add on and off

lib/commands/arm/webapp/webapp.config._js
LukaszStem: Good point. Will change.

PR #3164, opened 2016-09-15
ARM App gateway show command is fixed

MikhailTryakhov:
* Fixed app gateway show command format


MikhailTryakhov:
would be reworked in next PRs

PR #3168, opened 2016-09-16
switch to use file based token storage (enable creds sharing with new…

yugangw-msft:
* By default azure-cli will now save access tokens to ~/.azure/accessTokens.json for OSX and Window, like it does on Linux. When you install this version, please run `login` to re-establish the credentials. If you prefer old behaviors of using secure storage, you can turn on the env variable of `AZURE_USE_SECURE_TOKEN_STORAGE`


yugangw-msft:
//cc @mayurid

yugangw-msft:
fix azure/azure-cli#627

PR #3170, opened 2016-09-17
Add tests for 'vm set --new-os-disk-size'

huangpf:
* 
* 
*


huangpf:
@amarzavery This is the added tests for 'vm set --new-os-disk-size'

PR #3183, opened 2016-09-21
upgrade azure-common and azure-storage to resolve tough-cookie issue

amarzavery:
* upgrade azure-common and azure-storage to resolve tough-cookie issue


amarzavery:
@hasonmsft - Do you think upgrading to azure-storage 1.3.0 should be fine?

PR #3227, opened 2016-10-06
Linux on App Service

LukaszStem:
Changes:  …
Add: linux option when creating appserviceplans
Add: tests for linux option
Add: showing is linux in webapp config show


lib/commands/arm/appserviceplan/appserviceplan._js
```diff
@@ -53,6 +53,7 @@ exports.init = function (cli) {
     .option('-l --location <location>', $('the geographic region to create the App Service Plan, \n\t\t\t\t\t   locations found at https://azure.microsoft.com/en-us/status/'))
     .option('-k --sku <sku>', $('sku of the App Service Plan' + tierInfo.toString()))
     .option('-i --instances <instances>', $('the number of instances in the App Service Plan, Default is 1'))
+    .option('-x --islinux <islinux>', $('enables a linux App Service Plan, E.g --islinux true'))
```
amarzavery: this could be a boolean flag. Which when specified the value is true, else it is false. This makes things lot more simpler. You dont have to parse the value that the user provides (True, true, yes, y, etc.)
` .option('-x --islinux', $('enables a linux App Service Plan, E.g --islinux'))`

lib/commands/arm/appserviceplan/appserviceplan._js
```diff
@@ -87,9 +88,28 @@ exports.init = function (cli) {
       var envelope = { 
         location: location,
         name: name,
-	      sku: skuInfo
+        sku: skuInfo
       };
       var result;
+      if ('islinux' in options) {
```
amarzavery: you can simply say 
```
var customOptions = {};
customOptions.linuxWorkersEnabled = options.islinux;
```

No need to even check that whether islinux is present in options.

lib/commands/arm/appserviceplan/appserviceplan._js
```diff
@@ -87,9 +88,28 @@ exports.init = function (cli) {
...
+          log.info('Checking if this is a supported region');
+          var customOptions = {
+            linuxWorkersEnabled: true
+          };
+          try {
+            result = client.global.getSubscriptionGeoRegions(customOptions, _);
+          } finally {
+            progress.end();
+          }
+          if (result === undefined || result.length === 0) {
```
amarzavery: what if result is null ?

amarzavery: if (result === null || result === undefined || (result && result.length === 0))

lib/commands/arm/appserviceplan/appserviceplan._js
```diff
@@ -313,5 +334,25 @@ exports.init = function (cli) {
       }
     var skuInfo = {name: tier, tier: sku, family: tier.charAt(0)};
     return skuInfo;
-	}
+  }
+
+  function checkRegionForLinux(result, location) {
+    var locationEdit = (location.replace(/ /g, '')).toUpperCase();
+		var i;
```
yugangw-msft: nit: indent

LukaszStem: fixed

lib/commands/arm/appserviceplan/appserviceplan._js
LukaszStem: good point, fixed

lib/commands/arm/appserviceplan/appserviceplan._js
LukaszStem: To keep things consistent, I will keep it as --islinux true

PR #3240, opened 2016-10-10
Throw error when deployment create fails

vivsriaus:
Throw error when deployment create fails


lib/commands/arm/group/group.deployment._js
```diff
@@ -119,6 +118,9 @@ exports.init = function (cli) {
             displayDeployment(data, resourceGroup, true, log);
           }
         });
+        if (deployment.properties && deployment.properties.provisioningState !== 'Succeeded') {
```
amarzavery: What if it was Cancelled ? Will throwing an error, then make sense?

amarzavery: I am sure the service is sending 'Succeeded' in Pascalcase. However, it doesnt hurt to do case insensitive match.

PR #3241, opened 2016-10-10
Throw error when deployment fails

vivsriaus:
Throw error when deployment fails

### DIFF
--- a/lib/commands/arm/group/group.deployment._js
+++ b/lib/commands/arm/group/group.deployment._js
@@ -86,11 +86,9 @@ exports.init = function (cli) {
             displayDeploymentStatusMessage(cli, operationsIncludingNested, log, _);
 
             setTimeout(_, 5000);
-          } while (deployment.properties.provisioningState === 'Running' || deployment.properties.provisioningState === 'Accepted');
+          } while (utils.ignoreCaseEquals(deployment.properties.provisioningState, 'Running') || utils.ignoreCaseEquals(deployment.properties.provisioningState, 'Accepted'));
 
-          if (deployment.properties.provisioningState !== 'Succeeded') {
-            log.error('Deployment provisioning state was not successful');
-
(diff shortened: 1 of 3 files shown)


lib/commands/arm/group/group.deployment._js
```diff
@@ -86,11 +86,9 @@ exports.init = function (cli) {
             displayDeploymentStatusMessage(cli, operationsIncludingNested, log, _);
 
             setTimeout(_, 5000);
-          } while (deployment.properties.provisioningState === 'Running' || deployment.properties.provisioningState === 'Accepted');
-
-          if (deployment.properties.provisioningState !== 'Succeeded') {
-            log.error('Deployment provisioning state was not successful');
+          } while (utils.ignoreCaseEquals(deployment.properties.provisioningState, 'Running') || utils.ignoreCaseEquals(deployment.properties.provisioningState, 'Accepted'));
 
+          if (!utils.ignoreCaseEquals(deployment.properties.provisioningState, 'Succeeded')) {
```
amarzavery: Should you explicitly check for 'Failed', instead of checking for not Succeeded. This will make sure that you don't enter if the provisioningState was Cancelled.

lib/commands/arm/group/group.deployment._js
```diff
@@ -119,6 +118,9 @@ exports.init = function (cli) {
             displayDeployment(data, resourceGroup, true, log);
           }
         });
+        if (deployment.properties && utils.ignoreCaseEquals(deployment.properties.provisioningState, 'Failed')) {
+          throw new Error($('Deployment provisioning state was not successful.'));
```
TianoMS: why throw an error instead of log.error ?

vivsriaus: @TianoMS To make the command "Failed" instead of "OK" when deployment fails

TianoMS: Looks like it's the pattern in CLI. If that's the case, it should be fine.

PR #3247, opened 2016-10-11
Added appCommandLine option

LukaszStem:


package.json
```diff
@@ -13,7 +13,7 @@
     "Wang, Yugang <yugangw@microsoft.com>",
     "Chen, Hao <hach@microsoft.com>"
   ],
-  "version": "0.10.5",
+  "version": "0.10.6",
```
yugangw-msft: //cc:@amarzavery

PR #3253, opened 2016-10-12
Adding container commands

LukaszStem:
Really just modifying appsettings under the covers.


lib/commands/arm/webapp/webapp.config._js
```diff
@@ -548,19 +549,173 @@ exports.init = function (cli) {
...
+
+  container.command('set [resource-group] [name]')
+    .description($('Set webapp Container configurations (available only for linux webapps)'))
+    .usage('[options] <resource-group> <name>')
+    .option('-g --resource-group <resource-group>', $('the name of the resource group'))
+    .option('-n --name <name>', $('the name of the webapp'))
+    .option('-r --url <url>', $('the container registry server url'))
+    .option('-u --username <username>', $('the container registry server username'))
+    .option('-p --password <password>', $('the container registry server password'))
+    .option('-c --customimagename <customimagename>', $('the container custom image name and optionally the tag name'))
```
yugangw-msft: Should be `--custome-image-name`

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -548,19 +549,173 @@ exports.init = function (cli) {
...
+      var subscription = profile.current.getSubscription(options.subscription);
+      var client = webappUtils.createWebappManagementClient(subscription);
+      var progress = cli.interaction.progress($('Getting webapp configuration'));
+      var currentAppSettings;
+      try {
+        currentAppSettings = client.sites.listSiteAppSettings(resourceGroup, name, _);
+      } finally {
+        progress.end();
+      }
+      var appSettingsPrevious = currentAppSettings.properties;
```
yugangw-msft: can you just directly update the original copy?

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -548,19 +549,173 @@ exports.init = function (cli) {
...
+      var settingSplit;
+      var result;
+      var appSettingsProperties = {};
+      var actualAppSettings = {};
+
+      for (var attrname in appSettingsPrevious) { appSettingsProperties[attrname] = appSettingsPrevious[attrname]; }
```
yugangw-msft: This is not right as it will loop through all meta properties. Like i commented early, please try to use the original one, if for reasons you can't, then you should at least check `hasOwnProperty(key)`

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -548,19 +549,173 @@ exports.init = function (cli) {
...
+      } finally {
+        progress.end();
+      }
+      var appSettingsPrevious = currentAppSettings.properties;
+      var settingSplit;
+      var result;
+      var appSettingsProperties = {};
+      var actualAppSettings = {};
+    
+      for (var attrname in appSettingsPrevious) {
```
yugangw-msft: same issue here.

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -548,19 +549,173 @@ exports.init = function (cli) {
...
+      try {
+        currentAppSettings = client.sites.listSiteAppSettings(resourceGroup, name, _);
+      } finally {
+        progress.end();
+      }
+      var appSettingsPrevious = currentAppSettings.properties;
+      var settingSplit;
+      var result;
+      var appSettingsProperties = {};
+      var actualAppSettings = {};
```
yugangw-msft: Should you error out if no configurations are specified, or just skip the rest of the command?

test/commands/arm/webapp/arm.webapp.feature-tests.js
```diff
@@ -149,6 +151,27 @@ describe('arm', function () {
       });
     });
 
+    it('config container set should work', function (done) {
+      suite.execute('webapp config container set %s %s -u %s --json', groupName, sitename, username, function (result) {
```
yugangw-msft: Can we test with real container images?

lib/commands/arm/webapp/webapp.config._js
LukaszStem: fixed

lib/commands/arm/webapp/webapp.config._js
LukaszStem: Why is hasOwnProperty necessary if it's being overwritten anyway?

test/commands/arm/webapp/arm.webapp.feature-tests.js
LukaszStem: Within the testing framework we have here, no. External action is required to check if the registry exists. Unless there is a way to "ping" this registry.

lib/commands/arm/webapp/webapp.config._js
LukaszStem: Fixed

lib/commands/arm/webapp/webapp.config._js
LukaszStem: fixed

lib/commands/arm/webapp/webapp.config._js
LukaszStem: fixed

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -548,19 +549,170 @@ exports.init = function (cli) {
...
+      var progress = cli.interaction.progress($('Getting webapp configuration'));
+      var currentAppSettings;
+      try {
+        currentAppSettings = client.sites.listSiteAppSettings(resourceGroup, name, _);
+      } finally {
+        progress.end();
+      }
+      var result;
+   
+      if (currentAppSettings.properties.hasOwnProperty('DOCKER_REGISTRY_SERVER_URL'))
```
yugangw-msft: i don't believe you need to check beforehand, `delete` will be a non op if the property doesn't exist

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -548,19 +549,170 @@ exports.init = function (cli) {
...
+      }
+
+      try {
+        result = client.sites.getSite(resourceGroup, name, _);
+      } finally {
+        progress.end();
+      }
+      currentAppSettings.location = result.location;
+      try {
+        result = client.sites.updateSiteAppSettings(resourceGroup, name, currentAppSettings, _);
```
yugangw-msft: unless you need to show different progress, you should move this line be in the same try block with `client.sites.getSite`, this will save some lines of code

test/commands/arm/webapp/arm.webapp.feature-tests.js
yugangw-msft: I see, ensure you manually verified it then

lib/commands/arm/webapp/webapp.config._js
LukaszStem: fixed

lib/commands/arm/webapp/webapp.config._js
LukaszStem: fixed

yugangw-msft: not fixed. The code starts the progress for 'getting web configuration', and `.end(` it when it is done. For `update` the finally block also has `.end(`,  but has no matched `.interaction.progress(`. My feedback is you do 2 things in the same try block, or start a new progress for `update`.

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -548,19 +549,155 @@ exports.init = function (cli) {
...
+      }
+      delete currentAppSettings.properties['DOCKER_REGISTRY_SERVER_URL'];
+      delete currentAppSettings.properties['DOCKER_REGISTRY_SERVER_USERNAME'];
+      delete currentAppSettings.properties['DOCKER_REGISTRY_SERVER_PASSWORD'];
+      delete currentAppSettings.properties['DOCKER_CUSTOM_IMAGE_NAME'];
+      currentAppSettings.location = result.location;
+      try {
+        result = client.sites.updateSiteAppSettings(resourceGroup, name, currentAppSettings, _);
+      } finally {
+        progress.end();
```
yugangw-msft: same issue here.

LukaszStem: fixed

lib/commands/arm/webapp/webapp.config._js
LukaszStem: I believe it is now fixed. Have merged into 1 try block.

PR #3275, opened 2016-10-25
Fixed ARM network commands descriptions

MikhailTryakhov:
- Fixed ARM network commands descriptions


amarzavery:
>     .option('-s, --subscription [subscription]', $('the subscription identifier'))

Why are you changing `<subscription>` to `[subscription]` . subscription is a required parameter. We auto populate it in the cli for ease of use. However, semantically one needs to provide it.

---
Refers to: lib/commands/arm/network/network._js:191 in 8803574. [](commit_id = 880357496a719323b86747a621c133891744fe32, deletion_comment = False)

amarzavery:
@MikhailTryakhov - Can you respond to the above comment ?

MikhailTryakhov:
Hi @amarzavery sorry for delay.
Got it and would revert such changes.
That is not an urgent fix and need for generator mostly.

PR #3314, opened 2016-11-13
Add --force-update-tag support to VM Extension & update test

huangpf:
* Add --force-update-tag support to VM Extension & update test


lib/plugins.arm.json
```diff
@@ -38679,6 +38679,14 @@
                   "description": "Pick the highest minor version of the extension"
                 },
                 {
+                  "flags": "--force-update-tag <force-update-tag>",
```
balajikris: No short notation for this? something like a `-f` ?

balajikris:
@huangpf please update description with changelist. This will go into the release notes.

lib/plugins.arm.json
huangpf: By design.

Issue #3283, opened 2016-10-29
No way to enable and/or set custom probe for application gateway?


nneul:
0.10.6 / npm / linux

Unless I'm missing something, I can't seem to find any way to enable a custom probe on http-settings for an application gateway. I can create and manage the probe without issue, but I was expecting to find a "--custom-probe enabled" and "--probe-name name" set of options on http-settings set, but not seeing it. Clearly the cookie setting can be done, but not the probe unless it's somewhere else. 

```
nneul@skyhawk:~$ azure network application-gateway http-settings set --help
help:    Update http settings of an application gateway
help:    
help:    Usage: network application-gateway http-settings set [options] <resource-group> <gateway-name> <name>
help:    
help:    Options:
help:      -h, --help                                           output usage information
help:      -v, --verbose                                        use verbose output
help:      -vv                                                  more verbose with debug output
help:      --json                                               use json output
help:      -g, --resource-group <resource-group>                the name of the resource group
help:      -w, --gateway-name <gateway-name>                    the name of the application gateway
help:      -n, --name <name>                                    the name of the HTTP settings
help:      -p, --protocol <protocol>                            the protocol, valid value is [http]
help:      -o, --port <port>                                    the port, valid range is [0,65535]
help:      -c, --cookie-based-affinity <cookie-based-affinity>  enable or disable cookie based affinity, valid values are
help:           [Disabled,Enabled],
help:           default value is [Disabled]
help:      --nowait                                             does not wait for the operation to complete. Returns as soon as the intial request is received by the server.
help:      -s, --subscription <subscription>                    the subscription identifier
```

nneul:
Looks like it's just an attribute in the settings collection like cookieBasedAffinity - just needs to be an id. 

```    "backendHttpSettingsCollection":[
    {
      "name":"<settingsName>",
      "id": "/subscriptions/<subscriptionId>/../microsoft.network/applicationGateways/<gatewayName>/backendHttpSettingsList/<settingsName>",
      "location": "<location>",
      "tags": {"key": "value"},
      "etag": "W/\"00000000-0000-0000-0000-000000000000\"",

      "properties": {
        "port":"<backendPortNumber>",
        "protocol":"<backendProtocol>" ,
        "cookieBasedAffinity":"Enabled",
        "requestTimeout": "<requestTimeoutInSec>",
[...]

PR #3315, opened 2016-11-14
Fixed issue #3283

MikhailTryakhov:
* Fixed issue [#3283](https://github.com/Azure/azure-xplat-cli/issues/3283)
* Added an ability to associate app gw probe and http settings

### DIFF
--- a/lib/commands/arm/network/appGateway._js
+++ b/lib/commands/arm/network/appGateway._js
@@ -700,7 +700,7 @@ __.extend(AppGateways.prototype, {
     }
 
     var httpSettings = { name: httpSettingsName };
-    httpSettings = self._parseHttpSettings(httpSettings, options, true);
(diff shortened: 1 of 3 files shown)


nneul:
Any chance you could add support for probe 'name' as well since the probes are already directly associated with the app gateway when they are created?

MikhailTryakhov:
@nneul thatnks,
it's implemented too

amarzavery:
Installers of the next release (0.10.8) will be made available on December 14th. If you need this feature right now then you can download the tarball from the dev branch and install the cli via npm in the global location. Customers installing the CLI on linux, install it from the tarball we provide, hence the process by and large remains the same.

Please follow the instructions documented over [here](https://github.com/Azure/azure-xplat-cli#installation-from-a-particular-branch-of-this-repository) to install the latest bits from a particular branch.

PR #3338, opened 2016-11-23
Add VM Secrets Support

huangpf:
* Add VM Secrets Support


huangpf:
@amarzavery @balajikris 
Please help review (but don't merge yet). Thanks.

lib/commands/arm/vm/vmClient._js
```diff
@@ -2171,6 +2171,62 @@ __.extend(VMClient.prototype, {
...
+    };
+    if (options.certificateUrl) {
+      var certUrlObj = {
+        certificateUrl : options.certificateUrl,
+        certificateStore : options.certificateStore,
+      };
+      sourceVaultObj.vaultCertificates.push(certUrlObj);
+    }
+    vmResult.osProfile.secrets.push(sourceVaultObj);
+    virtualMachine.createOrUpdateVM(params.group, vmResult, false, _);
```
amarzavery: Is the command displaying correct output with and without --json switch? I am asking this because I do not see that code over here.

lib/commands/arm/vm/vm._js
```diff
@@ -931,4 +931,50 @@ exports.init = function (cli) {
...
+        sourceVaultId: sourceVaultId
+      };
+      vmClient.addSecret(params, options, _);
+    });
+
+  secret.description($('Commands to manage VM secrets'));
+  secret.command('delete [group] [name] [source-vault-id]')
+    .description($('Deletes a secret to the VM'))
+    .usage('[options] <group> <name> <source-vault-id>')
+    .option('-g, --group <group>', $('the resource group'))
```
amarzavery: --resource-group

lib/commands/arm/vm/vm._js
```diff
@@ -931,4 +931,50 @@ exports.init = function (cli) {
 
         vmClient.getVMImageDetails(params, options, _);
       });
+
+  var secret = vm.category('secret').description($('Commands to manage VM secrets'));
+  secret.command('add [group] [name] [source-vault-id]')
+    .description($('Adds a secret to the VM'))
+    .usage('[options] <group> <name> <source-vault-id>')
+    .option('-g, --group <group>', $('the resource group'))
```
huangpf: ok

lib/commands/arm/vm/vmClient._js
huangpf: Other commands have the same, so I think it should be, but I can double check.

huangpf: confirmed.

squillace:
@amarzavery is there a description of the behavior here? I need something to hang a document on. :-)

amarzavery:
@huangpf - Can you help @squillace with this?

huangpf:
Assuming users are familiar with the VM REST API (https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines), the new commands open up control on the following parameters in the VM profile:
 +    .option('-r, --source-vault-id <source-vault-id>', $('the source-vault-id'))
 +    .option('-c, --certificate-url <certificate-url>', $('certificate-url'))
 +    .option('-t, --certificate-store <certificate-store>', $('certificate-store'))

squillace:
Yes, but what HAPPENS? what, precisely, does the call **do**?

huangpf:
I think the call will happen to add a secret setup to the VM profile, which means that a secret resource will be installed to the VM physically. But I would expect the REST API documents this part as well, so that we don't have to duplicate.

PR #3342, opened 2016-11-23
Fix resource create and set commands to take in the passed in properties

vivsriaus:
* Fix resource create and set commands to take in the passed in properties


amarzavery:
@vivsriaus - Can you please add tests for these? Or improve assertions in the current tests?

lib/commands/arm/resource/resource._js
```diff
@@ -74,7 +74,7 @@ exports.init = function (cli) {
     .option('--subscription <subscription>', $('the subscription identifier'))
     .execute(function (resourceGroup, name, resourceType, properties, apiVersion, options, _) {
```
amarzavery: @vivsriaus 
.execute(function (resourceGroup, name, resourceType, **properties**, apiVersion, options, _) {

- properties is defined explicitly over here. Why are you taking this from options (options.properties) ?

Please add a test or tighten the current tests. I am not sure that this is still working.

vivsriaus:
@amarzavery Yes, tests are still pending. I was waiting for the node SDK PR to be merged, so I can add new tests with the new sdk

lib/commands/arm/resource/resource._js
vivsriaus: Added a test to ensure passed in properties is indeed honored

amarzavery:
@vivsriaus - resolve conflicts

test/commands/arm/resource/arm.resource-tests.js
```diff
@@ -65,13 +65,38 @@ describe('arm', function () {
...
+      });
+
       it('should work without switches', function (done) {
         var groupName = suite.generateId('xTestResource', createdGroups, suite.isMocked);
         var resourceName = suite.generateId('xTestGrpRes', createdResources, suite.isMocked);
         suite.execute('group create -n %s --location %s --json', groupName, testGroupLocation, function (result) {
           result.exitStatus.should.equal(0);
 
-          suite.execute('resource create %s %s %s %s %s --json', groupName, resourceName, 'Microsoft.Web/sites', testApiVersion, '{ "name": "' + resourceName + '", "siteMode": "Limited", "computeMode": "Shared" }', function (result) {
+          suite.execute('resource create %s %s %s %s -p %s --json', groupName, resourceName, 'Microsoft.Web/sites', testApiVersion, '{ "name": "' + resourceName + '", "siteMode": "Limited", "computeMode": "Shared" }', function (result) {
```
TianoMS: >-p %s [](start = 53, length = 5)

This test case seems to be testing "without switches". Do we want to add -p here?

lib/commands/arm/resource/resource._js
```diff
@@ -52,7 +52,7 @@ exports.init = function (cli) {
     .option('--subscription <subscription>', $('the subscription identifier'))
     .execute(function (resourceGroup, name, resourceType, apiVersion, options, _) {
       var values = getValuesForCommand(resourceGroup, name, resourceType, apiVersion, options);
-      resource.createResource(values.resourceGroup, values.resourceName, values.resourceType, values.apiVersion, values.properties, values.options, _);
+      resource.createResource(values.resourceGroup, values.resourceName, values.resourceType, values.apiVersion, options.properties, values.options, _);
```
TianoMS: >options.properties [](start = 113, length = 18)

Can we update "getValuesForCommand" so we can use values.properties?

PR #3343, opened 2016-11-24
fix get-serial-console issue 3266

ejarvi:
Extracts host from blob URI so that blob service can default to the correct endpoint suffix.


amarzavery:
@ejarvi - can you add some tests for this? How do we know that this is indeed the correct fix?

ejarvi:
Thanks Amar.


I think the question is --- Do we need to add a test, or do we need to add a test environment?


From what I have observed, the command was only broken within the national clouds<https://www.microsoft.com/en-us/trustcenter/CloudServices/NationalCloud>.   It also appears to have existing test coverage in the following tests:

  *   test/commands/arm/vm/arm.vm-stop-restart-tests.js<https://github.com/Azure/azure-xplat-cli/blob/bea341b53149246fb61b95883aa34b72a7e980f2/test/commands/arm/vm/arm.vm-stop-restart-tests.js>
  *   test/commands/arm/vm/arm.vm-create-tests.js<https://github.com/Azure/azure-xplat-cli/blob/f24bb28d9b94bf956af0e2c61ab8753434e92826/test/commands/arm/vm/arm.vm-create-tests.js>
  *   test/commands/arm/vm/arm.vm-premium-create-tests.js<https://github.com/Azure/azure-xplat-cli/blob/bea341b53149246fb61b95883aa34b72a7e980f2/test/commands/arm/vm/arm.vm-premium-create-tests.js>

If these tests are sufficient, then in national cloud environment I would expect these tests to fail prior to this change, and to succeed after this change.


Do you know if these tests are currently being run in the national cloud environments?  If not, that might be how we missed the bug the first time.





________________________________

amarzavery:
@ejarvi - Today is code freeze. Any update on adding a test for this? Have you manually tested this in the national cloud environment?

Issue #3339, opened 2016-11-23
`azure network application-gateway create` can't customize some parameters


zhongyi-zhang:
CLI Version: **0.10.2**
OS Type: Linux
Installation via:  npm

Mode: **ARM**

Environment: **AzureCloud**

Description:
`azure network application-gateway create` can't customize some parameters.

Steps to reproduce:
Can't find some parameters in help. And it returns warnings below.

Error stack trace:

warn:    Using default gateway ip name: ipConfig01
warn:    Using default frontend ip name: frontendIp01
warn:    Using default frontend port name: frontendPort01
warn:    Using default address pool name: pool01
warn:    Using default http settings name: httpSettings01
warn:    Using default http listener name: listener01
warn:    Using default request routing rule name: rule01

amarzavery:
@MikhailTryakhov @jtuliani - Please take a look at this issue.

MikhailTryakhov:
It's fixed and would be published in next release on December

PR #3345, opened 2016-11-27
Arm app gw create commands improved

MikhailTryakhov:
* Fixed issue [#3339](https://github.com/Azure/azure-xplat-cli/issues/3339)
* Added an ability to change default names for app gateway http-settings, http listener, frontend port, frontend ip, gateway IP config
* Fixed extra-logger and mistypes in commands


lib/commands/arm/network/appGateway._js
```diff
@@ -2066,7 +2065,7 @@ __.extend(AppGateways.prototype, {
       }
     }
     if (!options.gatewayIpName) {
-      options.gatewayIpName = 'ipConfig01';
+      options.gatewayIpName = constants.appGateway.gatewayIp.name;
```
balajikris: Nice!

PR #3349, opened 2016-11-28
[batch-beta-dev] Feature/batch template testing

theunrepentantgeek:
* Unit testing and associated fixes for batch template expansion

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -3,6 +3,7 @@
   "camelcase": true,
   "curly": false,
   "eqeqeq": false,
+  "esnext": true,
   "forin": false,
   "immed": true,
   "indent": 2,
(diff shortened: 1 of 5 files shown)


xingwu1:
Please change the title to add "[batch-beta-dev]".

theunrepentantgeek:
The CI build is failing with this err:

    /home/travis/build/Azure/azure-xplat-cli/lib/commands/batch/batch.templateUtils._js:656
      let updatedString = "";
      ^^^

But, my google-fu is failing me and I haven't found a solution - suggestions solicitated!

xingwu1:
The error said: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode.
So maybe we should to add "use strict" in the templateUtils._js?

PR #3350, opened 2016-11-28
Add OS and command diagnostics in User agent header

balajikris:
* Add OS info and command info to user agent string.
* Refactored code around telemetry.
* Added tests for the new scenarios.

### DIFF
--- a/lib/util/telemetry.js
+++ b/lib/util/telemetry.js
@@ -217,7 +217,7 @@ exports.currentCommand = function (command) {
   if (command && typeof command === 'object') {
     _currentCommand = command;
 
-    assert(_event, "telemetry._event is falsy when it wasn't expected.");
+    assert(_event, 'telemetry._event is falsy when it was not expected.');
     _event.commandName = command.fullName();
     _event.command = _filterCommand(_event.commandName, _rawCommand);
   }
--- a/lib/util/telemetry.js
+++ b/lib/util/telemetry.js
@@ -143,10 +143,6 @@ var _filterCommand = function (commandName, rawCommand) {
 var _stop = function (qosEvent) {
   if (qosEvent) {
     qosEvent.duration = Date.now() - qosEvent.startTime;
-    if (_currentCommand) {
-      qosEvent.commandName = _currentCommand.fullName();
-      qosEvent.command = _filterCommand(qosEvent.commandName, _rawCommand);
-    }
   }
 };
 
@@ -214,6 +210,8 @@ exports.init = function (isEnabled) {
 exports.currentCommand = function (command) {
   if (command && typeof command === 'object') {
     _currentCommand = command;
+    _event.commandName = command.fullName();
+    _event.command = _filterCommand(_event.commandName, _rawCommand);
   }
 };
 
@@ -263,3 +261,19 @@ exports.onFinish = function (callback) {
   }
 };
 
+exports.getUserAgentInfo = function () {
+  if (_event) {
+    return {
+      osType: _event.osType,
+      osVersion: _event.osVersion,
+      nodeVersion: _event.nodeVersion,
+      installationType: _event.installationType,
+      userId: _event.userId,
+      subscriptionId: _event['Azure.Subscription.Id'],
+      userType: _event.userType,
+      mode: _event.mode,
+      commandName: _event.commandName,
+      rawCommand: _rawCommand.slice(2, _rawCommand.length).join(' ')
+    };
+  }
+};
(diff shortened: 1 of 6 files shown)


balajikris:
@yugangw-msft @amarzavery : Please review when you get a chance.

balajikris:
ugh.. tests passed locally for me but are on the floor on the CI server. Likely a result of combination of telemetry being disabled on CI machines and my changes that run when telemetry is off, which would have otherwise not have run.

yugangw-msft:
@balajikris, in your local box, you can do `azure account clear` before run the test, this should repro the CI failure

balajikris:
Observations:


the failure stack indicates this:
arm.powerbiembedded-tests.js:19:13  ---> require('utils.js') --> require(telemetry) --> require(profile) --> require(adalAuth)
we are resolving require dependencies and hitting a failure in adalAuth.

I introduced "telemetry" in the dependency chain... previously utils.js did not require telemetry.js.

Let me try to fix this.

balajikris:
The issue is that this introduces a circular dependency. With my change, `utils` requires `telemetry`. however `telemetry` depends on `utils` indirectly [via `profile`, `adalAuth`].  

So, when i broke the dependency chain by making `adalAuth` not rely on `utils`, but rather use `utilsCore`, the failure moved down, to `environment` which seems to have the same problem. This makes me think `circular dependency` is indeed the issue and i need make `utils` not rely on `telemetry` (the one i introduced). 

Despite all this, how VS Test runner continues to pass all my tests, I may never understand.

balajikris:
That commit did get the tests started. Now it seems some individual command tests have failed. Investigating if it was due to the fact that the userAgent in request header was modified?

balajikris:
Note: This PR is now ready for review.

lib/util/telemetry.js
```diff
@@ -214,11 +216,23 @@ exports.init = function (isEnabled) {
...
   if (command && typeof command === 'object') {
     _currentCommand = command;
+
+    if (_event) {
+      _event.commandName = command.fullName();
+      _event.command = _filterCommand(_event.commandName, _rawCommand);
+
+      // update user agent info about the current command
+      userAgentCore.setCommandInfo(_event.commandName,
+        _rawCommand.slice(2, _rawCommand.length).join(' '));
```
yugangw-msft: The `_rawCommand.length`  can be omitted

lib/util/userAgentCore.js
```diff
@@ -0,0 +1,44 @@
...
+  _userAgentData.mode = mode;
+};
+
+exports.setUserAgentData = function (data) {
+  _userAgentData = data;
+};
+
+exports.getUserAgentData = function () {
+  if (_userAgentData) {
+    return _userAgentData;
```
yugangw-msft: You can just return `_userAgentData`; since the `else` condition will always return `undefined`.

lib/util/utils.js
```diff
@@ -47,10 +48,19 @@ exports.stringStartsWith = utilsCore.stringStartsWith;
 exports.pathExistsSync = utilsCore.pathExistsSync;
 
 var getUserAgent = exports.getUserAgent = function () {
-  var userAgent = util.format('WindowsAzureXplatCLI/%s', moduleVersion);
+  var userAgent = util.format('AzureXplatCLI/%s', moduleVersion);
   if (process.env.AZURE_HTTP_USER_AGENT) {
     userAgent += ';' + process.env.AZURE_HTTP_USER_AGENT;
```
yugangw-msft: We should return here to respect the environment variable; otherwise it is a breaking change

test/commands/cli.telemetry-test.js
```diff
@@ -81,6 +83,142 @@ describe('cli', function() {
...
+
+      // assert command name and parameters.
+      should(userAgentInfo).have.property('commandName').with.type('string').be.equal(givenCommand);
+      should(userAgentInfo).have.property('rawCommand').with.type('string').be.equal(expectedRawCommand);
+
+      done();
+    });
+
+    it('should construct user agent info object when telemetry is not enabled', function (done) {
+      var track = sandbox.spy(applicationInsights.client, 'track');
```
yugangw-msft: this should be moved to setup like `beforeEach`, so to avoid code dupes

lib/util/telemetry.js
balajikris: thanks for the tip. choosing to keep the verbose expression instead of brevity, because it is easier to read. `slice(start, end)` is a bit easier to read. Let me know if you feel strongly about it :-)

lib/util/utils.js
balajikris: Done -- thanks!

balajikris:
@yugangw-msft -- Thanks for taking a look. I've made the changes that were asked, or have answered to the comments left. I'd love it if you could take another quick look at it and sign off. Thanks much.

balajikris:
reviewed again with yugang, got a oral sign off to merge.

PR #3360, opened 2016-12-01
Change deployment polling logic.

TianoMS:
* 
* 
*

### DIFF
--- a/lib/commands/arm/group/group.deployment._js
+++ b/lib/commands/arm/group/group.deployment._js
@@ -77,7 +77,18 @@ exports.init = function (cli) {
       if (!options.nowait) {
         cli.interaction.withProgress($('Waiting for deployment to complete'),
         function (log, _) {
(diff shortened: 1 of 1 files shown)


lib/commands/arm/group/group.deployment._js
```diff
@@ -77,15 +77,27 @@ exports.init = function (cli) {
       if (!options.nowait) {
         cli.interaction.withProgress($('Waiting for deployment to complete'),
         function (log, _) {
+          // Poll deployment state and deployment operations with two phases. In phase one, poll every 5 seconds. Phase one
```
amarzavery: @TianoMS - Why do you need to manually poll? Autorest generated code does the polling for you. Any particular reason to do this?

vivsriaus: @amarzavery we need to manually poll for us to display verbose messages of where a deployment is (resources completed, in progress, failed, nested deployment status) when the deployment is running. This is not possible in auto rest today.

Issue #3302, opened 2016-11-07
humanOutputGenerator is not a function


jonathanong:
CLI Version: **0.10.7**
OS Type: Mac
Installation via:  npm

Mode: **ARM / ASM**

Description:
`webapp config appsettings set` throws error

Error stack trace:

```
2016-11-07T08:23:43.172Z:
{ TypeError: humanOutputGenerator is not a function
  <<< async stack >>>
    at Interactor.formatOutput (/Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/lib/util/interaction.js:307:7)
    at __2 (/Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:255:35)
  at __2 (/Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:252:35)
  <<< raw stack >>>
    at Interactor.formatOutput (/Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/lib/util/interaction.js:307:7)
    at __$__2 (/Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:255:35)
    at ___ (/Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:253:94)
    at ___ (/Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/node_modules/streamline/lib/callbacks/runtime.js:111:13)
    at /Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/node_modules/azure-arm-website/lib/operations/sites.js:13210:12
    at retryCallback (/Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/node_modules/ms-rest/lib/filters/exponentialRetryPolicyFilter.js:140:9)
    at handleRedirect (/Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/node_modules/ms-rest/lib/filters/redirectFilter.js:36:9)
    at /Users/jong/.nvm/versions/node/v6.9.1/lib/node_modules/azure-cli/lib/util/utils.js:587:7
[...]

PR #3364, opened 2016-12-02
Fixed cli.interaction in webapp config set

LukaszStem:
One line change, this cmd will no longer generate 
## error:   humanOutputGenerator is not a function

Please let me know if there is a more proper way of fixing this.


lib/commands/arm/webapp/webapp.config._js
```diff
@@ -252,7 +252,7 @@ exports.init = function (cli) {
             result = client.sites.createOrUpdateSiteConfig(resourceGroup, name, jsonObj, _);
             log.info('Webapp ' + name + ' configuration has been updated ');
           }
-          cli.interaction.formatOutput(result);
+          cli.interaction.formatOutput(result, function(){});
```
amarzavery: this is incorrect. The function cannot be empty. That function tells the CLI how the data in the result will be displayed to the user. An empty function will do nothing. It will surely not throw an error "humanOutputGenerator" is not a function. But, how useful is this empty function? Shouldn't you be printing the content of the result?

LukaszStem: I do this so I can process json when testing here https://github.com/Azure/azure-xplat-cli/blob/dev/test/commands/arm/webapp/arm.webapp.feature-tests.js#L119 . Or is this statement not needed to do that?

amarzavery: @LukaszStem I understand that. But as a customer who is executing the command, what is the output that he will see on the console. This function is used to display that output. Currently that function is empty so the user will not see anything. 

You should use this function to display the content of result object. Take a look at [this example](https://github.com/Azure/azure-xplat-cli/blob/dev/lib/commands/arm/cdnmanagement/cdnmanagement._js#L73).

LukaszStem:
For this command all I am showing that an update to the config either succeeded or failed and nothing more. I don't need to use result. Although I have updated it to check if the result is empty of null.

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -252,11 +252,15 @@ exports.init = function (cli) {
             result = client.sites.createOrUpdateSiteConfig(resourceGroup, name, jsonObj, _);
             log.info('Webapp ' + name + ' configuration has been updated ');
           }
-          cli.interaction.formatOutput(result);
-
+          
+          cli.interaction.formatOutput(result, function () {
+            if (!result || result.length === 0)
+              log.error($('API call did not return a valid result'));
```
balajikris: 1. what does the user see on success? if the user doesnt need to see anything on success, then this its alright. just checking to see if it was intentional.

2. this error message is too generic that it could be unhelpful to the end user, as well as to us to diagnose what went wrong. consider outputting a meaningful error message.

LukaszStem: This is what the user sees [`Capture of output`](https://s28.postimg.org/psbn0klv1/Capture.png)

The user gets the log.info messages here -> https://github.com/LukaszStem/azure-xplat-cli/blob/836cde3236ddd57650120b3cf99cb79982d561f5/lib/commands/arm/webapp/webapp.config._js#L247 

Proper errors are shown. It really shouldn't get to this if statement. I keep it there because of what's been stated in previous comments.

amarzavery:
https://github.com/Azure/azure-xplat-cli/issues/3302

PR #3372, opened 2016-12-06
Servicefabric fix update service command

msjeffreychen:
* fix update service command, instanceCount would not be udpated
* 
*


balajikris:
is there a bug number that this addresses?

msjeffreychen:
No, there is an email thread about this problem. If you would like, I could create a bug.

balajikris:
Ah, I just asked so I could get context to review the code. Could you please forward me the context or create a bug on this repo? Either one is fine with me, I'd like to understand the problem before looking at the fix :)
Thank you.

msjeffreychen:
Forwarded the email thread.

balajikris:
Thanks, Change looks alright to me. If its possible to add tests for this case, that'd be nice. If not, I'm not too concerned. Thanks.

msjeffreychen:
Please merge for us. Thanks.

PR #3378, opened 2016-12-07
URGENT fix with test config

MikhailTryakhov:


MikhailTryakhov:
@amarzavery 
decided that it would be better that @andrew-mestas  would fix tests in his merged PR and would publish fixes with tests config restored.

Really sorry guys that I've missed it.

PR #3379, opened 2016-12-07
testlist and version in nock recording fix

andrew-mestas:
* 
* Fixed broken test due to api version in nock recording and reverted testlist-arm.txt
*


balajikris:
if you're reverting some change, its usually better to revert the commit when possible than manually edit the file to revert the unintended change. That way the file's commit history will make it easy to understand what had happened.

PR #3384, opened 2016-12-08
[batch-beta-dev] Update Batch Shipyard integration to version 2.1.0

jasper-schneider:
* Update Batch Shipyard integration to version 2.1.0


lib/commands/batch/batch.batchShipyardUtils._js
```diff
@@ -107,7 +108,7 @@ batchShipyardUtils.invokeBatchShipyard = function(credentialsJson, configJson, p
 
   try {
     // Invoke Batch Shipyard
-    var command = 'python ' + batchShipyardPath + ' --configdir ' + batchShipyardConfigDir + ' ' + action;
+    var command = util.format('python %s %s --configdir %s', batchShipyardPath, action, batchShipyardConfigDir);
     cli.output.info('Invoking Batch Shipyard: ' + command);
```
darylmsft: what is the decision point between using util.format (above) and "+" here?

jasper-schneider: Basically, the command was looking ugly with multiple concats for variables and spaces, but the logging message seemed simple enough for one concat.

PR #3423, opened 2017-01-05
Adding import, export and Reset commands to Xplat CLI for Azure Redis Cache

alfantp:
* adding import, export and Reset commands to Xplat CLI for Azure Redis Cache
* 
*


balajikris:
Hey @alfantp : I'm unable to review this PR. The change mainly is in `rediscache._js`. 
However it appears there were whitespace changes to the entire file and github shows every line as changed. [See screenshot 1]
I'm able to ignore whitespace changes by adding `?w=1` to the diff but then Github doesn't let me add any comments in this view [See screenshot 2]. 
The actual change seems to be 300 lines worth but the diff as submitted shows 2000 lines. 

Could you separate out the whitespace changes if they're intentional? 

**Screenshot 1**
image

**Screenshot 2**
image

balajikris:
Hey @alfantp -
I couldn't leave inline comments (due to issue mentioned in above comment), but I went through the diff (ignoring whitespace). Some comments:

**File:** rediscache._js

1. _Lines 234, 318, 393:_ there is a verbose log here, is this intended for the end user? if not, should this be removed?

2. _Lines 265, 346, 423:_ casing for local variable (start with lower case)

3. _Lines 291, 371, 448:_ is variable 'result' used? its assigned once, but I don't see where it is used.
4. _Lines 302, 459:_ after invoking the export command, you check for a 200 and print a status. How do we handle reporting for other status codes? Does it fall back to an outer catch block that notifies the user the command failed.
6. _Lines 303, 379 , 460:_ Status reporting: take this message for example: "Import command successfully invoked for Redis Cache".  It mentions the command was invoked but it is not clear whether the command itself "completed" successfully or not. Try tweaking the message if you agree.
7. _Line 460_: copy paste error. the user facing message should be about "reset" instead of "export".
8. Weird diff in this file, which i mentioned above with screenshots. I think checking this in as-is would make git-history, git-annotate hard on this as it would like as if this change touched every line in this file. That said, I'm okay with it, if you are okay with it.

Please clarify about these. Otherwise, things look good.

alfantp:
Thanks for reviewing balajikris. Response to your comments inline:
1. Corrected
2. I removed the ones that are not used.
3. For the commands where i have added check for status 200, it is just to make sure the command invocation succeeded as it doesn't give any other response. For other status codes, the command would fail and show the error to user.
4. This is correct. The response only indicates whether command was invoked correctly. The command gets executed asynchronously.
5. Corrected
6. I had the spacing as a mistake in one of my commits, but in a following commit, i had fixed the casing.

balajikris:
Thanks! Changes look good to me 👍 

Regarding 6 - I don't see the fix to spacing. Github still shows me 2,053 lines changed for RedisCache._js

I'm okay with the functional changes, if you clarify with the space issue or are okay with it, I'll go ahead and merge after the test run succeeds.

alfantp:
I'm okay with this @balajikris . Thanks a lot for reviewing

balajikris:
Thanks @alfantp. Merged in.

PR #3427, opened 2017-01-06
Servicefabric

msjeffreychen:
* Add timeout to application type register command


msjeffreychen:
Thanks for reminder. Updated.

PR #3430, opened 2017-01-09
[batch-beta-dev] Sample and documentation improvements

theunrepentantgeek:

* Renaming sample folders to better reflect their content
* Modify data migration sample to use per-file task factory instead of parametric sweep
* Improvements to documentation to aid discoverability of features


Documentation/BatchDocumentation/readme.md
```diff
@@ -0,0 +1,56 @@
+# New Features for Azure Batch
```
matthchr: We already have a readme.md at the top level of the repo -- can we merge this with that?

Documentation/BatchDocumentation/readme.md
```diff
@@ -0,0 +1,56 @@
...
+azure batch file upload C:\data\**\*.png raw-images
+
+azure batch file upload /tmp/data/**/*.png raw-images
+```
+
+## [Output file upload](outputFiles.md)
+
+When adding a task, you can now declare a list of output files to be automatically uploaded to an Azure Storage container of your choice when the task completes.
+
+**Note:** This feature currently only available on Linux VMs.
```
matthchr: No longer true, it works on Windows now

Documentation/BatchDocumentation/samples/hello-world/create-pool-and-job-with-templates/readme.md
```diff
@@ -0,0 +1,78 @@
+# Azure Batch Pool/Job Template
+
+This sample shows how to create a pool, and run a parametric sweep job on it, using *parameterised* templates.
```
matthchr: Do we have a write up somewhere of what a parameterised template is and why you might want to use it?

Documentation/BatchDocumentation/readme.md
theunrepentantgeek: I want to have a specific landing page that lists the new features, rather than just a directory full of markdown files - goal is to improve discovery of the features so they are more likely to be picked up and used.

That said, I'll have a look at the top level readme and see if there's a good way to link in there too.

theunrepentantgeek: The top level readme already seemed to have some good links, so I didn't need to do much.

Documentation/BatchDocumentation/readme.md
theunrepentantgeek: Thanks; I've removed the note.

Documentation/BatchDocumentation/samples/hello-world/create-pool-and-job-with-templates/readme.md
theunrepentantgeek: The docs I wrote for application templates touch on this, but I'm not sure Xing has written similar for ARM-style templates.

Documentation/BatchDocumentation/samples/hello-world/taskPerFile-appTemplate/readme.md
```diff
@@ -38,7 +39,7 @@ To create your job, run the following command:
 azure batch job create --json-file job.json
 ```
 
-This job uses an **autopool** which will be automatically deleted once the job finishes. 
+This job uses an **autopool** which will be automatically deleted once the job finishes. If the job stalls before all the tasks are complete, you may need to use the [Azure portal](https://portal.azure.com) and manually delete the pool to ensure you're not billed unnecessarily.
```
jasper-schneider: What exactly do you mean by "stalls" here?
Tasks aren't being scheduled?
Tasks hanging?
Something else?
All of the above?

I'm just trying to figure out if there's a more precise way to word it, but I have no suggestions, since I'm not 100% clear on what the error condition here is.

Documentation/BatchDocumentation/readme.md
```diff
@@ -0,0 +1,54 @@
...
+azure batch file upload /tmp/data/**/*.png raw-images
+```
+
+## [Output file upload](outputFiles.md)
+
+When adding a task, you can now declare a list of output files to be automatically uploaded to an Azure Storage container of your choice when the task completes.
+
+## [Templates](templates.md)
+
+Templates allow pools and jobs to be defined in parameterised json files with a format inspired by ARM templates.
```
theunrepentantgeek: Aligning on 'murican spelling is probably wise for consistencies sake.

Documentation/BatchDocumentation/readme.md
```diff
@@ -0,0 +1,54 @@
...
+
+## [Shipyard (Docker)](shipyard.md)
+
+An integration with Batch Shipyard to allow you to provision Batch compute nodes with Docker containers and to schedule Docker workloads. For more information on Batch Shipyard, see its [GitHub page](https://github.com/azure/batch-shipyard).
+
+**Note:** This feature is only available on Linux VMs.
+
+## [Application Templates](application-templates.md)
+
+Application templates provides a way to partition the details of a job into two parts.
```
theunrepentantgeek: "Application Templates" is indeed a noun representing the entire feature - but we can't expect users to know that. I'll amend the wording.

Documentation/BatchDocumentation/readme.md
```diff
@@ -0,0 +1,54 @@
...
+
+All of the details about how the job should be processed are moved into the **application template**, creating a reusable definition that is independent of a particular account. Application templates are parameterized to allow the processing to be customized without requiring modification of the template itself.
+
+## [Package managers](packages.md)
+
+Integration with existing 3rd party package managers to streamline the installation of applications. Cuirently the following package managers are supported:
```
theunrepentantgeek: Fixed.

Documentation/BatchDocumentation/readme.md
```diff
@@ -0,0 +1,54 @@
...
+
+* Chocolately - for Windows
```
theunrepentantgeek: Wow, I've done that **everywhere**.

Documentation/BatchDocumentation/readme.md
matthchr: My point was that the top level readme and your readme seem to do something very similar. I am not sure why we need both was my main comment

Documentation/BatchDocumentation/samples/ffmpeg/readme.md
```diff
@@ -1,54 +1,116 @@
...
+**You are billed for your Azure Batch pools, so don't forget to delete it through the [Azure portal](https://portal.azure.com) when you're done.** 
+
+## Using a parametric sweep for processing
 
 ### Upload files
-Run command `azure batch file upload <path> <group>` on a folder containing media files (`*.wav`) which are named with numerically increasing names with `sample` prefix (i.e. `sample1.wav`, `sample2.wave`, `sample3.wav`, etc).
 
-### Create a job with parametric sweep tasks
-Edit the `job.parameters.json` file to supply parameters to the template. If you want to configure other options of the job, such as the the pool id, you can look in the `job.json` parameters section to see what options are available.
+Upload your WAV media files by running this command on a folder containing media files (`*.wav`) which are named with numerically increasing names with `sample` prefix (i.e. `sample1.wav`, `sample2.wave`, `sample3.wav`, etc):
```
theunrepentantgeek: Rephrased.

Documentation/BatchDocumentation/samples/ffmpeg/readme.md
```diff
@@ -1,54 +1,116 @@
...
 
-### Create a job with a task generated per input file
-Edit the `job.parameters.json` file to supply parameters to the template. If you want to configure other options of the job, such as the the pool id, you can look in the `job.perFile.json` parameters section to see what options are available.
+Upload your WAV media files by running this command on a folder containing media files (*.wav):
+
+``` bash
+azure batch file upload <path> <group>
+```
+
+With this sample there is no restriction on file naming aside from all the files have same extension (`*.wav`).
```
theunrepentantgeek: Rephrased as "Unlike the sample using parametric sweep, there's no requirement for your filenames to confirm to a specific pattern."

Documentation/BatchDocumentation/samples/hello-world/create-pool-and-job-with-templates/readme.md
theunrepentantgeek: I'm going to install a 'murican spellchecker into Visual Studio Code. 😆

Documentation/BatchDocumentation/samples/hello-world/create-pool-and-job-with-templates/readme.md
```diff
@@ -0,0 +1,78 @@
...
+You can also use the [Azure portal](https://portal.azure.com) or [Batch Explorer](https://github.com/Azure/azure-batch-samples/tree/master/CSharp/BatchExplorer) for monitoring.
+
+## Structure of the sample
+
+### pool.json
+The file `pool.json` contains a template for defining a new pool with three parameters defined:
+
+| Parameter | Required | Description                                                                                                           |
+| --------- | -------- | --------------------------------------------------------------------------------------------------------------------- |
+| poolId    | Required | The id of the Azure Batch pool                                                                                        |
```
theunrepentantgeek: No random newlines - just lots of "spaces" so that the | bars | of the table align vertically.

Documentation/BatchDocumentation/samples/hello-world/create-pool-and-job-with-templates/readme.md
```diff
@@ -0,0 +1,78 @@
...
+
+The file `pool.parameters.json` provides values for the parameters defined in `pool.json`. You will need to provide a value for `poolId` before pool creation will succeed. If you do not want to use the default values for `vmCount` or `vmSize`, add values for those parameters to this file before creating the pool.
+
+### job.json
+
+The file `job.json` contains a template for a new job with four parameters defined:
+
+| Parameter | Required  | Description                                   |
+| --------- | --------- | --------------------------------------------- |
+| jobId     | Mandatory | The id of Azure Batch job                     |
```
theunrepentantgeek: I want to have all the tables with the heading "Required" and values "Mandatory" vs "Optional". Thanks for catching the inconsistencies.

Documentation/BatchDocumentation/samples/hello-world/taskPerFile-appTemplate/readme.md
theunrepentantgeek: Trying to capture a situation that happened to me - had a job that didn't run to completion, and of course the auto-pool was sitting there, idle, because it's kept until the job finishes. Wanted to flag to an end user that they may want to manually clean it up if this happens to them.

Documentation/BatchDocumentation/readme.md
theunrepentantgeek: Last time I looked at the readme in the root of the repo, it hadn't been customized for NCJ, so I didn't realize there was duplication. That said, I'm thinking of the users who might be browsing through the GitHub repo and looking at the samples folder; having a readme right in front of them is useful, even if the same information is available elsewhere.

matthchr: The readme at the root of the repo has definitely been customized for NCJ. I just want to avoid duplication. I would rather see us have a doc that links to there, or edit the upper level readme to link to this doc, something to avoid 2 competing documents which both are trying to describe "what the NCJ features are and where you can learn more about them" which I feel both your readme and the top level readme are trying to do.

theunrepentantgeek: I see your point. Will try to work out something.

Documentation/BatchDocumentation/samples/hello-world/taskPerFile-appTemplate/readme.md
jasper-schneider: Ok, so the main issue is that if a job doesn't reach the completed state, then the pool will hang around and you'll get billed.  I think "stalls" is a little vague and we can be a little more explicit with Batch service terminology.  How about something like this? (Also, I just saw that you need to insert a "be" between "automatically" and "deleted")

This job uses an autopool which will automatically be deleted once the job reaches the Completed state. If the job does not reach the Completed state (ex: tasks are unable to run because of scheduling errors or errors downloading files from storage), then you may need to use the ...

Documentation/BatchDocumentation/readme.md
theunrepentantgeek: I've moved all of the content about the features up into the landing page for the repo branch, leaving just a summary of links in the directory for the feature docs.

Documentation/BatchDocumentation/samples/hello-world/taskPerFile-appTemplate/readme.md
theunrepentantgeek: I like your alternative phrasing, I'll use it here and elsewhere. 

I'll also change "will be automatically deleted" to "will automatically be deleted" - while the former is valid, the later is clearer.

Issue #3435, opened 2017-01-11
azure webapp config hostnames list fails


setevoy2:
CLI Version: **azure webapp config hostnames list**
OS Type: Linux (Arch Linux)
Installation via:  npm

Mode: both - **ARM / ASM**

Environment: **AzureCloud**

Description:
`azure webapp config hostnames list` returns "Cannot read property 'length' of undefined".

Steps to reproduce:
1) Run `azure login`
2) Then run `azure webapp config hostnames list -g resgrname -n webappname`

Error stack trace:

```$ cat /home/setevoy/.azure/azure.err
2017-01-11T10:08:42.124Z:
TypeError: Cannot read property 'length' of undefined
    at /usr/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:598:22
    at Interactor.formatOutput (/usr/lib/node_modules/azure-cli/lib/util/interaction.js:320:7)
    at __$__9 (/usr/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:596:27)
    at __tryCatch (/usr/lib/node_modules/azure-cli/node_modules/streamline/lib/callbacks/runtime.js:150:4)
    at ___ (/usr/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:594:177)
    at ___ (/usr/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:594:110)
    at __tryCatch (/usr/lib/node_modules/azure-cli/node_modules/streamline/lib/callbacks/runtime.js:150:4)
    at ___ (/usr/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:594:65)
    at __$__9 (/usr/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:594:31)
    at __tryCatch (/usr/lib/node_modules/azure-cli/node_modules/streamline/lib/callbacks/runtime.js:150:4)
TypeError: Cannot read property 'length' of undefined
    at /usr/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:598:22
    at Interactor.formatOutput (/usr/lib/node_modules/azure-cli/lib/util/interaction.js:320:7)
    at __$__9 (/usr/lib/node_modules/azure-cli/lib/commands/arm/webapp/webapp.config.js:596:27)
[...]

PR #3436, opened 2017-01-11
Fixed hostnames list "undefined" error

LukaszStem:
Regarding issue [#3435](https://github.com/Azure/azure-xplat-cli/issues/3435)


lib/commands/arm/webapp/webapp.config._js
```diff
@@ -595,7 +595,10 @@ exports.init = function (cli) {
       }
       cli.interaction.formatOutput(result, function (data) {
         data = data.value;
-        if (data.length > 0) {
+        if (data === undefined) {
+          log.info($('No web app hostname bindings defined.'));
+        }
+        else if (data.length > 0) {
```
balajikris: 1. Can you help me understand why the `data.length > 0` check was insufficient to make this flow take the existing `else branch` at line 608 ?

2. if you need this additional check, you could do `if (data && data.length > 0)` instead of an additional branch. It has some advantages: 

- your current condition doesn't check all `falsey` values (e.g: `null`). 
- the control flow in my suggestion is clear. "do this operation if data is valid, otherwise, log a message" instead of a multi way branch.

balajikris:
ping @LukaszStem can you please address the above PR comments?

lib/commands/arm/webapp/webapp.config._js
```diff
@@ -595,16 +600,15 @@ exports.init = function (cli) {
       }
       cli.interaction.formatOutput(result, function (data) {
         data = data.value;
-        if (data.length > 0) {
+        if (data === undefined || data.length <= 0 || !data) {
```
balajikris: nit: i think `!data` is sufficient to cover the `undefined` case as well.

PR #3452, opened 2017-01-25
Dev dns import hot fix

MikhailTryakhov:
* Fixed issue with TXT records import even if values contain record types (MX, TXT, etc)


MikhailTryakhov:
Hi @balajikris 
this code block is ready for review.
Please let me know your thoughts

PR #3454, opened 2017-01-26
Change setup authoring to install 32 or 64 bit Node based on target cpu arch

### LINKED ISSUE #3451: Install 64 bit node.exe on 64 bit windows
CLI Version: 0.10.8
OS Type: Win
Installation via:  installer 

Mode: ALL

Environment: ALL

Description:
The installed node.exe platform is x86 on x64 machine. It's better to install 32-bit on 32-bit windows and 64-bit on 64-bit windows?

balajikris:
* This change modifies the setup authoring to install 64bit NodeJs on
64bit CPUs and prefer 32bit NodeJs on 32bit CPUs. The CLI authoring
harvests the clone of source repo and deploys it to target directory. We
want to opt out of that process for authoring node.exe. We then author
the appropriate bitness node.exe based on VersionNT64 conditional check.
* Fixes #3451

### DIFF
--- a/tools/windows/scripts/prepareRepoClone.cmd
+++ b/tools/windows/scripts/prepareRepoClone.cmd
@@ -75,7 +75,7 @@ mkdir %TEMP_AUX_REPO%
 echo Downloading node x86 and x64...
 pushd %TEMP_AUX_REPO%
 mkdir %X86%
-pushd %x86%
+pushd %X86%
 curl -o node.exe %NODE_X86_DOWNLOAD_URL%
 if %errorlevel% neq 0 goto ERROR
 popd
--- a/tools/windows/scripts/prepareRepoClone.cmd
+++ b/tools/windows/scripts/prepareRepoClone.cmd
@@ -18,7 +18,8 @@ set PATH=%PATH%;"C:\Program Files (x86)\Git\bin;"
 
 pushd %~dp0..\
 
-set NODE_DOWNLOAD_URL=http://nodejs.org/dist/v%NODE_VERSION%/win-x86/node.exe
+set NODE_X86_DOWNLOAD_URL=http://nodejs.org/dist/v%NODE_VERSION%/win-x86/node.exe
+set NODE_X64_DOWNLOAD_URL=http://nodejs.org/dist/v%NODE_VERSION%/win-x64/node.exe
 set NPM_DOWNLOAD_URL=http://nodejs.org/dist/npm/npm-%NPM_VERSION%.zip
 
 echo Cleaning previous build artifacts...
@@ -28,7 +29,12 @@ if exist %OUTPUT_FOLDER% rmdir /s /q %OUTPUT_FOLDER%
 mkdir %OUTPUT_FOLDER%
 
 set TEMP_REPO_FOLDER=zcli
+set TEMP_AUX_REPO_FOLDER=zNode
+set X86=x86
+set X64=x64
 set TEMP_REPO=%HOMEDRIVE%%HOMEPATH%\%TEMP_REPO_FOLDER%
+set TEMP_AUX_REPO=%HOMEDRIVE%%HOMEPATH%\%TEMP_AUX_REPO_FOLDER%
+
 if not exist %TEMP_REPO% goto CLONE_REPO
 
 echo Temporary clone of the repo already exists. Removing it...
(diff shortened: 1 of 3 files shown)


balajikris:
@joeloff - please review if you get a chance.

balajikris:
@yugangw-msft - please review if you get a chance.

joeloff:
Looks good. I'd eventually consider having WiX generate stable GUIDs for your components instead of authoring them directly. It will also ensure component reference counts work if you ever need to author shared components in your MSIs to support side-by-side installs.

tools/windows/scripts/prepareRepoClone.cmd
```diff
@@ -53,17 +59,56 @@ if %errorlevel% geq 8 (
...
 unzip -q npm.zip
 if %errorlevel% neq 0 goto ERROR
 del npm.zip
 popd
 
+echo.
+echo Copying node into bin directory...
+:: We copy node.exe to bin directory to compile streamline files and other tasks.
+:: During clean up phase, we delete it from bin directory. The reason for doing this is because the
+:: entire source repo is harvested during set up authoring. We want to author node by hand after determining architecture type (32 or 64 bit).
```
yugangw-msft: I don't quite understand what does `author node by hand` mean.  We do package both version under `zNode`, right?

tools/windows/scripts/prepareRepoClone.cmd
```diff
@@ -53,17 +59,56 @@ if %errorlevel% geq 8 (
...
+    goto ERROR
+)
+popd
+
+:DOWNLOAD_TASK
+mkdir %TEMP_AUX_REPO%
+echo Downloading node x86 and x64...
+pushd %TEMP_AUX_REPO%
+mkdir %X86%
+pushd %x86%
```
yugangw-msft: nit, should be `X`86

tools/windows/scripts/prepareRepoClone.cmd
```diff
@@ -53,17 +59,56 @@ if %errorlevel% geq 8 (
...
 del npm.zip
 popd
 
+echo.
+echo Copying node into bin directory...
+:: We copy node.exe to bin directory to compile streamline files and other tasks.
+:: During clean up phase, we delete it from bin directory. The reason for doing this is because the
+:: entire source repo is harvested during set up authoring. We want to author node by hand after determining architecture type (32 or 64 bit).
+pushd %TEMP_AUX_REPO%\%X86%
+robocopy . %TEMP_REPO%\bin /NFL /NDL /NJH /NJS
```
yugangw-msft: you can just use the regular `copy` command since we only deal with a single node.exe, right?

tools/windows/scripts/prepareRepoClone.cmd
balajikris: setup authoring for xplat cli is fully auto-generated by a heat task (authoring is not hand-written). However, you could not exclude files under the heat task, so you cannot choose 32bit or 64bit version of node.exe conditionally. So, what we do is, take node.exe outside of cli's source root and write setup authoring to place the right version in the right location (authoring is hand-written and not auto generated). This is the new code in product.wxs. This is what I tried to explain in the above comment. Let me know if I should make it more clear.

to your question of whether we package both into the installer, yes, absolutely. That is how at install time, we choose which one to install.

tools/windows/scripts/prepareRepoClone.cmd
balajikris: I can, but does it make a difference?

balajikris:
To expand joeloff's comment: the suggestion was to remove Guid attributes in product.wxs and let WiX autogenerate them. This was for existing code. For the code I introduced, I still need to author explicit Guids. Failure to do so will result in the following warning, albeit it is technically harmless.

~~~
E:\repos\azure-xplat-cli\tools\windows\Product.wxs(128): warning LGHT1137: Component/@Id='node_x64' has a @Guid value '{B38FA75C-C8B8-596F-B858-B3C F2CDBA133}' that duplicates another component in this package. This is not officially supported by Windows Installer but works as long as all compo nents have mutually-exclusive conditions. It is recommended to give each component its own unique GUID. [E:\repos\azure-xplat-cli\tools\windows\azu re-cli.wixproj] 
~~~

PR #3455, opened 2017-01-27
Update for Storage commands

hasonmsft:
* Updated azure-storage module to 2.0.0
* Supported large block blob

Besides the Changelog, we'd like to request update the ReadMe to suggest user to: 
* Use 64-bit Windows for large block blob transfer
* Install 64-bit node on 64-bit Linux machine

Please let me know if this doable.

### DIFF
--- a/lib/util/storage.util._js
+++ b/lib/util/storage.util._js
@@ -292,8 +292,10 @@ StorageUtil.BlobSnapshotDeletion = {
 
 StorageUtil.MaxPolicyCount = 5;
 
+// 100 MB X 50000 blocks
+StorageUtil.MaxBlockBlobSize = 50000 * 100 * 1024 * 1024; 
+
 // 4 MB X 50000 blocks
-StorageUtil.MaxBlockBlobSize = 50000 * 4 * 1024 * 1024; 
 StorageUtil.MaxAppendBlobSize = 50000 * 4 * 1024 * 1024;
 
 // 1 TB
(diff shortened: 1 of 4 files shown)


lib/util/storage.util._js
```diff
@@ -292,8 +292,10 @@ StorageUtil.BlobSnapshotDeletion = {
 
 StorageUtil.MaxPolicyCount = 5;
 
+// 100 MB X 50000 blocks
+StorageUtil.MaxBlockBlobSize = 50000 * 100 * 1024 * 1024;
```
balajikris: would this affect code running on 32bit machines?

balajikris:
>  we'd like to request update the ReadMe to suggest user to

I'm thinking what's the best place to do this would be. For one, it has to be in the change log for this release, but then where would people find this information if they are on a later build than the next one.
Are you talking about the ReadMe.MD file? -- That would be for the repo, perhaps not the best place for the customer to know how to use the product.

Given this, I'm wondering if this information should be baked into the product itself, in the `help text` for this command? That would be rather straight forward to implement and the customer can easily find it when they are running into the problem or about to run into it. Thoughts?

/cc: @amarzavery

amarzavery:
@hasonmsft  - Please implement the feedback given by @balajikris .

balajikris:
@hasonmsft : given the codefreeze for next release is almost here, we can think about implementing the feedback in a separate PR, can you please get the tests passing so that we can get this in?

lib/util/storage.util._js
hasonmsft: It works fine. The only issue is that the available RAM is not enough for large block blob upload.

hasonmsft:
@balajikris , @amarzavery , thanks for your valuable feedbacks, the help text is updated accordingly.

PR #3460, opened 2017-02-01
[batch-beta-dev] Update the sample json and readme.

xingwu1:


Documentation/BatchDocumentation/samples/ffmpeg/job.perFile.json
```diff
@@ -1,8 +1,7 @@
 {
     "parameters": {
-        "dataGroup": {
+        "inputFileGroup": {
             "type": "string",
-            "defaultValue": "convert_data",
```
jasper-schneider: minor: What is the harm of having a default value?

Documentation/BatchDocumentation/samples/ffmpeg/readme.md
```diff
@@ -50,7 +50,7 @@ Modify `job.parameters.json` to supply parameters to the template. If you want t
...
 | -------------------- | --------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| poolId               | Mandatory | Name of the Azure Batch Pool to use for processing. <br/> Must match the pool you created earlier.                                                                            |                                                                                                                                                                        |
+| poolId               | Optional | Name of the Azure Batch Pool to use for processing. <br/> Must match the pool you created earlier. Default pool id is ffmpeg-pool.                                                                           |                                                                                                                                                                        |
```
jasper-schneider: minor: In other docs, the wording pattern was "Defaults to ffmpeg-pool".
ex: https://github.com/Azure/azure-xplat-cli/tree/batch-beta-dev/Documentation/BatchDocumentation/samples/blender

Documentation/BatchDocumentation/samples/mpi/job.json
```diff
@@ -11,7 +11,7 @@
             "type": "string",
             "defaultValue": "mpi",
             "metadata": {
-                "description": "The auto-storage group where the input data is stored"
+                "description": "The auto-storage group where the MSMPI program and dependencies files are stored"
```
jasper-schneider: Official branding seems to use a hyphen in MS-MPI: https://msdn.microsoft.com/en-us/library/bb524831(v=vs.85).aspx

Also, the wording is a little awkward.  How about this?
"The auto-storage group where the MS-MPI program and its dependencies are stored"

Documentation/BatchDocumentation/samples/mpi/readme.md
```diff
@@ -11,31 +11,37 @@ This samples shows how to use `MS-MPI` to run MPI work.
 
 You must have an Azure Batch account set up with a linked Azure Storage account.
 
-You will need a zip file containing a compiled version of the `MPIHelloWorld.exe` application and its dependencies. You may want a release build of `MPIHelloWorld.exe` to reduce the number of separate files.
+You will need an MS-MPI program for the multi-instance task to execute. We provide the [MPIHelloWorld sample project](https://github.com/Azure/azure-batch-samples/tree/master/CSharp/ArticleProjects/MultiInstanceTasks/MPIHelloWorld) for you to compile and use as your MS-MPI program. You may want a release build of `MPIHelloWorld.exe` to reduce the number of separate files.
```
jasper-schneider: grammar: "to compile and to use".  You're supposed to be consistent with the use of infinitives.

jasper-schneider: Also, while you're changing this, can you also change the last sentence?  Why bring up a point if we're going to be so vague about it?
How about we just use the text from Marsh's article?
Build a Release version of MPIHelloWorld.exe so that you don't have to include any additional dependencies (for example, msvcp140d.dll or vcruntime140d.dll) in your application package.

Documentation/BatchDocumentation/samples/mpi/readme.md
```diff
@@ -57,13 +63,13 @@ azure batch pool create --template pool.json --parameters <your settings JSON fi
...
-Upload sample files from a folder:
+Upload the `MPIHelloWorld.exe` application and its dependencies from a folder:
 
 ```bash
-azure batch file upload <path> <group>
+azure batch file upload <path> mpi
 ```
 
-Run this command in a folder containing your sample files. The parametric sweep expects the files to be named `sample1.wav`, `sample2.wav`, `sample3.wav` and so on - each with the prefix `sample` and an increasing index number. It's important for correct operation of the parametric sweep that your files are sequentially numbered with no gaps.
+You can use other name for file group name, just to make sure to update `inputFileGroup` parameter value in next section with same name.
```
jasper-schneider: I might tweak it to something like this?  Maybe other people can chime in on better wording.
`mpi` is the default value of the `inputFileGroup` parameter in the job template. If you upload your files to a different file group, be sure to provide this value for the `inputFileGroup` parameter when creating your job (see the next section).

Documentation/BatchDocumentation/samples/ffmpeg/job.perFile.json
theunrepentantgeek: Some of the other samples have a default pool name. (Not saying this one has to have a default, just noting that it's not inconsistent.)

Documentation/BatchDocumentation/samples/mpi/readme.md
theunrepentantgeek: How about "We suggest you use the MPIHelloWorld sample project."

Documentation/BatchDocumentation/samples/mpi/readme.md
theunrepentantgeek: Good catch on this; I missed it when I modified the sample to use `taskPerFile`.

Documentation/BatchDocumentation/samples/ffmpeg/job.perFile.json
xingwu1: We don't mention the default file group in the azure batch file upload command. I am fine with the default pool name.

Documentation/BatchDocumentation/samples/ffmpeg/readme.md
xingwu1: resolved.

Documentation/BatchDocumentation/samples/mpi/job.json
xingwu1: resolve.

Documentation/BatchDocumentation/samples/mpi/readme.md
xingwu1: resolve.

Documentation/BatchDocumentation/samples/mpi/readme.md
theunrepentantgeek: How about 

"You can use another name for the file group as long as you update the `inputFileGroup` parameter value in the next step to match."

xingwu1: resolve.

Documentation/BatchDocumentation/samples/ocr/job.json
```diff
@@ -45,7 +45,7 @@
                         {
                             "source": {
                                 "fileGroup": "[parameters('inputFileGroup')]",
-                                "prefix": "{fileName}.pdf"
+                                "prefix": "{fileNameWithoutExtension}.pdf"
                             }
                         }
```
theunrepentantgeek: When reviewing one of the other samples, Ivan noticed that this approach would give an odd error message if someone uploaded a non-pdf file; I ended up using `{fileName}` instead of `{fileNameWithoutExtension}.pdf` so that the errors would be more obvious.

xingwu1: so it should be {fileName} instead of {fileName}.pdf

PR #3475, opened 2017-02-06
Replaced existing vnet/subnet commands by generated version

MikhailTryakhov:
This version is a demo of existing CLI network services replacement by generated version


MikhailTryakhov:
Hi @balajikris @amarzavery 
our existing commands replacement is ready for merge.
Please review and let me know your thoughts

MikhailTryakhov:
@balajikris @amarzavery 
The same thing with [#3481](https://github.com/Azure/azure-xplat-cli/pull/3481),  [#3497](https://github.com/Azure/azure-xplat-cli/pull/3497), [#3498](https://github.com/Azure/azure-xplat-cli/pull/3498), [#3506](https://github.com/Azure/azure-xplat-cli/pull/3506), [#3523](https://github.com/Azure/azure-xplat-cli/pull/3523)

lib/commands/arm/network/virtualNetworks._js
```diff
@@ -0,0 +1,319 @@
...
+          result = null;
+        } else {
+          throw e;
+        }
+      } finally {
+        progress.end();
+      }
+      if (result) {
+        throw new Error(util.format($('virtual network with name "%s" already exists in the resource group "%s"'), name, resourceGroup));
+      }
```
balajikris: same comment as in the other file, rename the variable or the logic so as to not have `if(result) throw` pattern

lib/commands/arm/network/virtualNetworks._js
```diff
@@ -0,0 +1,319 @@
...
+        parameters.location = options.location;
+      }
+      if(!parameters.addressSpace) {
+        parameters.addressSpace = {};
+      }
+      if(options.addressPrefixes) {
+        parameters.addressSpace.addressPrefixes = [];
+        parameters.addressSpace.addressPrefixes = parameters.addressSpace.addressPrefixes.concat(options.addressPrefixes.split(','));
+      } else if(useDefaults) {
+        parameters.addressSpace.addressPrefixes = '10.0.0.0/8'.split(',');
```
balajikris: `'10.0.0.0/8'.split(',')` wont split anything. are you simply trying to create an array with this constant? If so, use array creation syntax, if not, this is a bug.

lib/commands/arm/network/virtualNetworks._js
```diff
@@ -0,0 +1,319 @@
...
+      }
+      if(!parameters.dhcpOptions) {
+        parameters.dhcpOptions = {};
+      }
+      if(options.dnsServers) {
+        if(!utils.argHasValue(options.dnsServers)) {
+          delete parameters.dhcpOptions.dnsServers;
```
balajikris: same comment as the other file, looks like you're invoking `removeEmptyObjects` in line 108. so you don't need to do this?

lib/commands/arm/network/virtualNetworks._js
```diff
@@ -0,0 +1,319 @@
...
+      }
+      if(options.addressPrefixes) {
+        if(result && result.addressSpace && result.addressSpace.addressPrefixes) {
+          parameters.addressSpace.addressPrefixes = result.addressSpace.addressPrefixes;
+        } else {
+          parameters.addressSpace.addressPrefixes = [];
+        }
+        parameters.addressSpace.addressPrefixes = parameters.addressSpace.addressPrefixes.concat(options.addressPrefixes.split(','));
+      } else if(useDefaults) {
+        parameters.addressSpace.addressPrefixes = '10.0.0.0/8'.split(',');
```
balajikris: same comment as above. there is no comma separated value in the constant to split it by commas.

lib/commands/arm/network/virtualNetwork-subnets._js
```diff
@@ -0,0 +1,365 @@
...
+          cli.output.warn($('--route-table-name parameter will be ignored because --route-table-id and --route-table-name are mutually exclusive'));
+        }
+        if(!utils.argHasValue(options.routeTableId)) {
+          delete parameters.routeTable;
+        } else {
+          parameters.routeTable.id = options.routeTableId;
+        }
+      } else if(options.routeTableName) {
+        if(!utils.argHasValue(options.routeTableName)) {
+          delete parameters.routeTable;
```
MikhailTryakhov: This is an action to remove related route table. It we call use option ```route-table-name```  with value  ``` "" ``` - ```!utils.argHasValue(options.routeTableName)``` statement would be true and we will have to remove route table parameter from the subnet

MikhailTryakhov:
Hi @evgenyagafonchikov 
please take part in the discussion

lib/commands/arm/network/virtualNetworks._js
MikhailTryakhov: wow

lib/commands/arm/network/virtualNetwork-subnets._js
MikhailTryakhov: answered above

lib/commands/arm/network/virtualNetworks._js
MikhailTryakhov: it's a hardcoded value :) it should be moved on separated variable.
but split still would be here in case if we would add multiple addresses here.
```split``` function separates address prefixes array

File: lib/commands/arm/network/virtualNetworks._js (hunk 3 above)
Comment (MikhailTryakhov) [inline lib/commands/arm/network/virtualNetworks._js]: the same as above

File: lib/commands/arm/network/virtualNetworks._js (hunk 2 above)
Comment (MikhailTryakhov) [inline lib/commands/arm/network/virtualNetworks._js]: the ame as below

File: lib/commands/arm/network/virtualNetwork-subnets._js (hunk 5 above)
Comment (MikhailTryakhov) [inline lib/commands/arm/network/virtualNetwork-subnets._js]: That's OK for ```set``` command, but can be reworked in ```create```

File: lib/commands/arm/network/virtualNetwork-subnets._js [hunk 6]
```diff
@@ -0,0 +1,365 @@
...
+        } else {
+          var idContainerrouteTable = networkManagementClient.routeTables.get(resourceGroup, options.routeTableName, _);
+          parameters.routeTable.id = idContainerrouteTable.id;
+        }
+      }
+
+
+      generatorUtils.removeEmptyObjects(parameters);
```
EvgenyAgafonchikov: Delete calls above are related to the situation like this https://github.com/Azure/azure-xplat-cli/pull/3475#discussion_r103859231

This call performs cleanup of objects that were created but were not filled with properties. E.g. if we created dhcpOptions but it still has no dnsServers to fill them dhcpOptions would be deleted before sending the object to server.

This call is related to the fact that some operations fail on server side if there are empty objects.
Alternative way is skipping of creating of empty objects if there is no options that create properties of this object. However this may lead to long `if` statements like
```
if(!parameters.object && !options.x && !options.y && !options.z) {
  parameters.object = {};
}
```
as some objects contain more than one property exposed via command options

balajikris: Thanks, It makes sense to iterate over and delete. I'm only concerned about using both individual deletes.

lib/commands/arm/network/virtualNetwork-subnets._js
balajikris: @MikhailTryakhov Thanks for explaining. I understand the need to delete, could you just use ` generatorUtils.removeEmptyObjects` to delete these or do you have to check them individually and delete like you do here?

I'm okay with it if thats your only option, but asking in order to see if this code could be made simple.

lib/commands/arm/network/virtualNetworks._js
balajikris: split if you have comma separated addresses, if you have only one hardcoded address do `foo = [10.0.0.0/8]`. I dont understand why you'd need to call split, if you know you this is a hard coded constant with only one address.

lib/commands/arm/network/virtualNetwork-subnets._js
EvgenyAgafonchikov: Adding small sample for clarification.
We created subnet by calling:
`network vnet subnet create <some params> --route-table-name myRouteTable`
Then we decided that there is no need in route table for this subnet and removing it using
`network vnet subnet set --route-table-name` command

So `argHasValue` in this case is for tracking that option `--route-table-name` is present by without argument.

`generatorUtils.removeEmptyObjects` is used to delete empty containers (if any) because of absent option key and value both.

This should be used in `set` command only (however doesn't break `create` workflow). Generator update to remove this form `create` and keep in `set` is near to completion then PR would be updated with this change.

balajikris: Thanks for explaining. Don't worry too much if it's hard to do. This is quite alright.

MikhailTryakhov:
Hi @balajikris 
could you please review upddated functionality in 3 first PRs?

lib/commands/arm/network/VirtualNetworks-subnets._js
```diff
@@ -0,0 +1,356 @@
...
+        } else {
+          throw e;
+        }
+      } finally {
+        progress.end();
+      }
+
+      if (!subnet) {
+        throw new Error(util.format($('subnet with name "%s" not found in the resource group "%s"'), name, resourceGroup));
+      }
```
balajikris: lines 149-165 is repeated for all commands. For e.g, see 65-81 in the above `create` command. essentially you can extract this common code into a helper function that looks something like: `checkAndThrowIfSubnetExists()` or `checkAndThrowIfSubnetDoesNotExist()`. 

If the validation logic refactored out, your command body will be cleaner. 

I'll leave it up to you as to whether or not you want to do it.

balajikris:
Hey @MikhailTryakhov - Changes look good to me. Thanks for addressing comments. I have one additional comment from this round of reviewing, which is to extract common code out in command impl. Its upto you whether you would like to do that clean up or not. Let me know what you prefer and I'll merge after that.

MikhailTryakhov:
Thanks @balajikris 
I assume that current implementation is completely the same as in handwritten code, so assume there is no need to change. Let's start merging :)

balajikris:
@MikhailTryakhov : sounds good. do you want to regenerate and update all the other PRs as well?

MikhailTryakhov:
@balajikris I'm in progress regenerating PRs. Would notify you in each ready for review code block

Issue #3444, opened 2017-01-18
ARM: When keyvault reference is specified in parameters, arm group deployment create shouldn't prompt for missing parameter


vivsriaus:
CLI Version: **0.9.X**
OS Type: Mac / Win / Linux (Ubuntu, RedHat, etc.)
Installation via:  npm / brew / win-or-mac installer / docker / github repo

Mode: **ARM / ASM**

Environment: **AzureCloud/Azure China Cloud/US Government/Blackforest/Azure Stack**

Description:
`azure vm show` command doesn't output VM name.

Steps to reproduce:
1) Run `azure vm create ...`
2) Then run `azure vm show ...`

Error stack trace:

**Please paste the content of ~/.azure/azure.err or C:\Users\username\\.azure\azure.err over here**

If the issue is w.r.t `authentication` then please set `AZURE_ADAL_LOGGING_ENABLED=1` and then run the `azure login` command again. 
Please paste the verbose logs over here. (Make sure to delete the password before pasting the contents).

PR #3488, opened 2017-02-09
Do not prompt for parameter value if a valid keyvault reference is in…

vivsriaus:
…cluded

* Do not prompt for parameter value if a valid keyvault reference is included
* Issue [here](https://github.com/Azure/azure-xplat-cli/issues/3444)


lib/commands/arm/group/groupUtils._js
```diff
@@ -464,11 +464,20 @@ function missingParameters(userParameters, options, subscription, _) {
...
+        }
+        return utils.stringIsNullOrEmpty(secretName) || utils.stringIsNullOrEmpty(keyVaultId);
+      } else {
         return !userParameters[parameter].value;
-      } catch(e){
-        return true;
       }
+    } catch (e) {
+      return true;
+    }
```
amarzavery: @vivsriaus If there is at least one test to validate the bug fix that would really help. It will help prevent regression and will be a sample to the customer.

vivsriaus: @amarzavery Agree on both. Added a test. Thanks!

test/data/arm-deployment-template-keyvault.json
```diff
@@ -0,0 +1,45 @@
+﻿{
+  "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "hostingPlanName": {
+      "type": "string"
```
TianoMS: >string [](start = 15, length = 6)

secureString?

TianoMS:
Change loos good. Just one small issue.

amarzavery:
@vivsriaus - Please address the issue pointed by @TianoMS . We are gearing for a new npm only release for cli. This thing can go in and solve customer problems.

test/data/arm-deployment-template-keyvault.json
vivsriaus: hostingPlanName is not a password. The intention of this test was to show that I can use keyvault reference in params (secure or not) to successfully deploy without prompting.

vivsriaus:
@amarzavery Addressed with a comment above. :) When is the new npm only release planned?

amarzavery:
02/22 when the npm release happens.

PR #3509, opened 2017-02-18
Update omelette for fish shell's completion

buzztaiki:
* Add auto completion for fish shell


balajikris:
@buzztaiki : Nice 👍

buzztaiki:
Thank you for merging.

> Did you get a chance to test if completion works after this change?

Some completion has been worked with zsh and bash on my laptop.

PR #3517, opened 2017-02-22
changes for release-0.10.10

balajikris:
updating package versions and changelog.


tools/windows/Product.wxs
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
 
-  <?define ProductVersion="0.10.9" ?>
+  <?define ProductVersion="0.10.10" ?>
```
balajikris: Though this is an npm only release, I'm updating this file anyway to keep it upto date, because the next installer package release version should be `0.10.11`

PR #3520, opened 2017-02-23
Add mac hash to user agent header

balajikris:
* add a SHA-256 hash of macAddress to userAgent header, per VS Telemetry standard.

### DIFF
--- a/lib/util/telemetry.js
+++ b/lib/util/telemetry.js
@@ -345,10 +345,11 @@ var _getMacAddressHash = function () {
 var _getMacOfFirstInterface = function (interfaces) {
   var nic;
   var validMacPattern = /^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$/;
+  var skipPattern = /^(0{2}[:-]){5}(0{2})$/;
   for (var key in interfaces) {
     if (interfaces.hasOwnProperty(key)) {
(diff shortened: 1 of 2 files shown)


lib/util/telemetry.js
```diff
@@ -297,7 +303,63 @@ function constructUserAgentData() {
...
   }
 }
+
+
+var _getMacAddressHash = function () {
+  var networkInterfaces = os.networkInterfaces();
+  var interfaceKeys = Object.keys(networkInterfaces);
+
+  var nicPattern;
+  if (os.platform().startsWith('win')) {
```
carlhoerberg: `os.platform()` does not return a string, it's an object!

balajikris: Not sure where you're seeing its an `object`.

But looking at https://nodejs.org/api/os.html#os_os_platform
~~~
os.platform()
Added in: v0.5.0
Returns: <string>

Currently possible values are:
'aix'
'darwin'
'freebsd'
'linux'
'openbsd'
'sunos'
'win32'
~~~

carlhoerberg:
It's making azure-cli unusable, crashes at every command. See issue #3546

balajikris:
yep, noticed the bug after i responded to this comment. I've responded to the bug now. Check it out and let me know if it worked.

PR #3528, opened 2017-03-01
CloudConsole: Support to establish profile and token cache from raw token

yugangw-msft:
This is for xplat to be used by cloud console hosted inside a browser. Portal will ping xplat routinely with tokens for xplat to establish profile and token cache, so users can execute cli commands


package.json
```diff
@@ -78,6 +78,7 @@
     "caller-id": "0.1.x",
     "colors": "1.1.2",
     "commander": "1.0.4",
+    "date-utils": "1.2.21",
```
yugangw-msft: ADAL-Node is already using it, and now we are also using it.

lib/util/profile/account.js
```diff
@@ -203,6 +209,69 @@ Account.prototype._getArmClient = function (credentials) {
         this._env.resourceManagerEndpointUrl);
 };
 
+Account.prototype._getSubscriptionInCloudConsole = function (callback) {
```
balajikris: consider adding some comments to this function as to what it is doing with the tokenCache and subscriptions list.  code looks good, but logic seems a bit involved without explaining comments.

yugangw-msft: done

PR #3552, opened 2017-04-05
login:allow accesstokens.json to be configurable through env var

yugangw-msft:
Doing for cloud console, that they like to redirect to /tmpfs (in memort) for better security


balajikris:
@azuresdkci  restest this please

PR #3554, opened 2017-04-06
update user agent logic for cli

balajikris:
1. use xplat cli's own unique user agent filter instead of mixing with
the rdfe runtime's filter. this was the rootcause of some bugs on how it
operates alongside of ms-rest runtime's own user agent filter as the cli
relies on ms-rest runtime to make requests. xplat cli's user agent
filter now correctly puts it UserAgent header in the request.

2. update ms-rest and ms-rest-azure runtime dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "Zavery, Amar <amzavery@microsoft.com>",
     "Wang, Yugang <yugangw@microsoft.com>",
     "Chen, Hao <hach@microsoft.com>",
-    "Krishan, Balaji <balajikris@microsoft.com>"
+    "Krishnan, Balaji <balajik@microsoft.com>"
   ],
   "version": "0.10.12",
   "description": "Microsoft Azure Cross Platform Command Line tool",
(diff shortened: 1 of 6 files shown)


balajikris:
1. manually tested that both arm/asm commands work fine with the change.
1. added test for user agent string
1. manually verified that the user agent string is what we expect in arm mode
1. under asm mode i didn't see user agent string in request header with or without this change. I may be missing something here but I verified that the request/response for asm commands were same with or without this change.

package.json
```diff
@@ -89,8 +89,8 @@
     "jwt-decode": "^2.1.0",
     "kuduscript": "1.0.13",
     "moment": "^2.8.0",
-    "ms-rest": "^1.12.0",
-    "ms-rest-azure": "^1.12.0",
+    "ms-rest": "^1.15.7",
+    "ms-rest-azure": "^1.15.7",
```
balajikris: keeping the upgrade minimal here because
1. regardless of what the runtime sets in UA, Xplat now takes care of setting the right thing, so run time version doesn't matter. That said, we know 15.5 had bugs, and 15.7 fixed those.
2. i feel the runtime is still churning a bit, given that we're so close to this release of xplat-cli, we should keep the churn minimal here and can upgrade to 2.0 for the next release. we get no added benefit of upgrading now anyway.

PR #3555, opened 2017-04-06
Update User Agent string and add tests

balajikris:
remove the sdk signature added with previous PR and add more tests to user agent filter


balajikris:
@yugangw-msft - here is the follow up 2 line delete as discussed

PR #3562, opened 2017-04-13
update query parameters in failing datalake tests as they changed wit…

balajikris:
…h 1.0.2-preview version of the package.


balajikris:
Thanks : @amarzavery @MikhailTryakhov @yugangw-msft
FYI, @begoldsm

begoldsm:
@balajikris thank you for this! I thought that the CLI was explicitly using the 1.0.1 version of the package (instead of >=). In the future I will continue the pattern you have, where we explicitly require a version to prevent these types of test failures.

Thanks again!

PR #3564, opened 2017-04-18
adding code for cred management with win token manager

amarzavery:


balajikris:
Is this built with the CI or are you just checking in source for a binary we depend on and we never build this until we make new changes?

PR #3567, opened 2017-04-18
VPN Gateway codebase is replaced with generated code

number213:
* VPN Gateway codebase is replaced with generated code
* Duplicating shorthand `f` in `vpn-gateway create` is fixed
* Expired certificate for testing VPN GW is replaced with a new one
* Test recordings are updated


test/testlist-arm.txt
```diff
@@ -1,3 +1,105 @@
-
+commands/arm/powerbiembedded/arm.powerbiembedded-tests.js
```
balajikris: why touch this file at all? If you're adding new network tests I'd agree, but you seem to be touching unrelated features. My guess is you're working off of a stale branch, commented tests to work around some issue and are now uncommenting it. Just drop those commits instead, otherwise it pollutes history for this file.

balajikris: at the least, squash some commits. That would ensure history for `testlist-arm.txt` is not polluted.

number213:
Sorry for the mess.
Commits are squashed, Travis build passed.

balajikris:
Thanks @number213 -- Merged!

PR #3568, opened 2017-04-18
Replacing nic code with autogenerated version

EvgenyAgafonchikov:


EvgenyAgafonchikov:
@MikhailTryakhov was replaced on this project so he could have no time to review these changes

MikhailTryakhov:
@balajikris that's true I even  have no laptop as well :) that's generated code without any new solutions, assume it's correct

PR #3570, opened 2017-04-19
Fix confirm string when remove nodes from Batch pool.

xingwu1:


lib/commands/batch/batch.pool._js
```diff
@@ -1055,7 +1055,7 @@ exports.init = function(cli) {
     }
     nodeList = interaction.promptIfNotGiven($('Nodes list: '), nodeList, _);
 
-    var tips = util.format($('Removing nodes from pool %s'), poolId);
+    var tips = util.format($('Removing node(s) \'%s\' from pool %s'), nodeList, poolId);
```
annatisch: Does the 'nodeList' here need the .join(', ') that it has in the formatting below?

xingwu1: No, nodeList is string.

balajikris:
hey @xingwu1 @annatisch 
quick question here. This PR is against `Azure:dev` and I wanted to check if it was intentional. I usually see you guys commit to `Azure:batch-beta` or `Azure:batch-beta-dev`. So, I'm curious if this was really meant to be against `Azure:dev`. ?

xingwu1:
@balajikris Yes, this is intentional. The batch-beta is using for some experiential features. However, this fix is  applied to official shipped cli v1.

xingwu1:
@balajikris or @amarzavery, could you merge this change?

PR #3576, opened 2017-04-26
Replacing local gateways with autogenerated code

EvgenyAgafonchikov:


balajikris:
@EvgenyAgafonchikov - have you tested this change manually? The reason i ask is because the next release is for [//Build](https://build.microsoft.com/), so we need to ensure its a high quality release. Another option is since this is just a re-write that doesn't add new features, we can merge this after we branch off for build. Let me know.

EvgenyAgafonchikov:
Yes, I've performed CRUD operations testing in several combinations for create/set.
This code is supposed to be just replacing of existing functionality with generated code to make adding of new features easier in future.

PR #3583, opened 2017-05-10
Integrate from master to dev as part of release-0.10.13

balajikris:


balajikris:
The appveyor build runs on merge commits as well, and fails, every time.

PR #3585, opened 2017-05-15
Update getobjectsbyobjectId call to use APi version 1.6 from 1.6-internal

darshanhs90:
This PR is the nodejs fix for the following issue https://github.com/Azure/azure-powershell/issues/3407
This also records the new sessions records for all the role tests and uncomments the role tests in the testlist.
Also contains the fix for the role-tests which had obsolete code as it used to call out the query method which was changed to queryForList(almost a year ago :-))
* 
* 
*


balajikris:
@darshanhs90 -Please get CI passing. I'll take a look once CI is all green.

darshanhs90:
@balajikris It is failing before running the tests and giving the error 
Error: ENOTEMPTY: directory not empty, rmdir '/home/travis/build/Azure/azure-xplat-cli/node_modules/azure-graph/node_modules/ms-rest'
at Error (native)
azure-graph@2.1.0-preview { [Error: ENOTEMPTY: directory not empty, rmdir '/home/travis/build/Azure/azure-xplat-cli/node_modules/azure-graph/node_modules/ms-rest']
Any advice on what needs to be done from my side to resolve this?

balajikris:
1. Please note that the current state of CI is green for the project (verify any closed PR to dev). 
2. One way to test that for sure is to drop the version of the package you're touching to the previous one and you'll see the CI go green.
3. which implies something isn't right with the latest version - i'm not sure what's wrong but nested node_modules folder looks suspicious to me. _node_modules should have been_ flattened at the same level, not nested. but that's just my guess. please dig deeper to find out what's wrong here.
4. if your investigation reveals somethings incorrect with our build infrastructure or if the build is broken, we'll investigate.

darshanhs90:
@balajikris When i reverted the version,the build fails due to the earlier version calling the API version 1.6-internal.So in order to fix this i need to update the version to 2.1.0-preview which was generated using your help.Hence when i make this change the build is failing in the **npm install** step itself.
The same **npm install** works on my local repo and doesnt complain like the issue earlier
So am not sure on how to go about this?Am i missing updating any other file ?

balajikris:
can you pull the latest from Azure:dev and retry once? I upgraded travis CI to use latest nodejs and npm.

darshanhs90:
@balajikris Now the npm install works,but however the npm test fails for a reason unrelated to my change.
**FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory**

balajikris:
I'm trying to find a more recent version of npm that'll work safely. My previous fix was to grab the latest - looks like latest-release (not preview) can have bugs like this. Instead of me and you juggling PRs to find a version of npm that'll work safely, lets do it another way.

1. see my PR https://github.com/Azure/azure-xplat-cli/pull/3588
2. incorporate that commit into this PR by doing "git cherry-pick 69293c13bb94f91e7ba11266af28d83e06458fe8" // which is the commit id from my PR.
3. If the travis CI build passes, we're good.
4. If it fails, we'll try falling back to slightly older versions of npm. (See https://semver.npmjs.com/)

I couldn't find a easier way to fix this, without involving your PR. somehow an older version of npm wasn't working well with your recent package -- not sure why.

balajikris:
FWIW I think the GC error would be a transient one. Simply because travis successfully ran tests on my PR which upgraded npm. This means if we just re-ran your PR tests might have turned up green. But it was a good catch, lets see if this falling back works [thought we might never be sure if the bug still exists in this version of npm we chose.] 💭

darshanhs90:
CI failed even after i cherrypicked your commit,with the same reason

balajikris:
Just saw that - its nodejs which runs into OOM, not npm 😢 

can you try this :

In that same file we were editing ".travis.yml", edit this line https://github.com/Azure/azure-xplat-cli/blob/dev/.travis.yml#L4 .
change it from "6" to "6.10.0"

I'll monitor it for failures.

balajikris:
what I don't understand is, it did run all tests and pass when I made the upgrade to use this version. I didn't add any extra tests but it ran the whole suite and passed. Somehow your PR runs into it. My next suggestion would be to drop your 1 commit that adds/modifies tests to see if you have something in your tests that's causing it to allocate resources endlessly and OOM.

balajikris:
Notice my other PR (https://github.com/Azure/azure-xplat-cli/pull/3588) from where we cherry-picked also is green. I don't think its a version of nodejs or npm that has an issue. Could be something we're doing in this PR that makes it OOM. I think CI will fail again, even with "6.10.0". If so, we should start by dropping your changes to tests and see if the existing suite passes with your product changes.

balajikris:
I'm still not clear why this would work now. though I am more than happy with "6.10.0" as that is the supported version for this project.

darshanhs90:
with 6.10.0 it passed :-)

balajikris:
The only plausible answer is, "6.10.3" the version installed when we said "6", has a bug, which the new tests you added/touched are hitting. That would explain why the existing suite always passed on my PRs. However your PR would always hit that bug. Sounds very remote, but that is the only reasonable explanation I can think of.

I'm okay with the npm version and nodejs version we settled here. If there are other commits which you want to drop (as we tried a lot back and forth), feel free to clean up the commits. I'll review and we can get to merging from there.

darshanhs90:
no these were the only changes

lib/commands/arm/role/roleAssignments._js
```diff
@@ -36,8 +36,13 @@ underscore.extend(RoleAssignments.prototype, {
   queryAssignmentsForList: function (principal, scope, roleName, roleId, shouldExpandPrincipalGroups, shouldIncludeClassicAdmins, cli, subscription, _) {
     var assignments;
     var objectType = {};
+    var principalId;
     var shouldRetrieveObjectType = shouldIncludeClassicAdmins;
-    var principalId = adUtils.getObjectId(principal, this.graphClient, false, shouldRetrieveObjectType, objectType, _);
+
+    if(principal !== null){
```
balajikris: Nit: Format document

balajikris:
Okay, I'm seeing continued OOM failures (this https://github.com/Azure/azure-xplat-cli/pull/3590 just failed and I restarted the build). Like I suspected all along, its in the tests that came along with this PR. A typical scenario is the tests are holding onto something for too long and not releasing it (incorrect clean up phase) - a standard thing could be that a recording file is held in memory too long.

Note that V8 enforces a limit of 1.8 G regardless of the machine's capacity. See [here](http://stackoverflow.com/questions/38558989/node-js-heap-out-of-memory). 

I'll monitor the queue closely and if disabling the test files that were added in this PR fixes the OOM issue, then we atleast know for sure that these tests are holding onto something. I'm beginning to wonder why they remained commented until you uncommented them in this PR.

balajikris:
/ cc : @mayurid

yugangw-msft:
@balajikris , for now you can split the "npm test" step into 2 sub steps of `npm unit` and `npm unit-arm` in the `.travis.yml`, which should ease the memory pressure. You can revert it back after you figure out the cause, likely we had too much test recordings loaded in the process and not released till process shutdown

balajikris:
Will try that as well, thanks @yugangw-msft !

darshanhs90:
not sure of the reason why they were commented out,since the test cases were outdated and didnt have any session records.I went ahead of fixing those and uncommented them

balajikris:
For what it's worth - rolling back the changes to testlist-arm.txt makes the CI go green several times in a row. To unblock build and all the other contributors, I rolled back your changes to `testlist-arm.txt` in this PR https://github.com/Azure/azure-xplat-cli/pull/3592 and filed an issue to re-enable these tests https://github.com/Azure/azure-xplat-cli/issues/3593.

darshanhs90:
okay,i'll look into it and figure out which test cases are causing the issues

PR #3592, opened 2017-05-18
Disable tests that are causing OOM failures in NodeJs

balajikris:


balajikris:
**_Observations:_**

1. Tests run into OOM in a dummy PR with no changes.

PR #3596, opened 2017-05-23
check OOM issue

darshanhs90:


balajikris:
hey @darshanhs90 - thanks for taking a stab at this. I think I might have fixed this issue. I saw this pop up in another PR today and spent some time with it here: https://github.com/Azure/azure-xplat-cli/pull/3600

My suggestions to you to re-enable your tests:
1. simply re-enable them in `testlist-arm.txt`, send a PR, reset the build atleast 3 times and ensure they are always green.
2. If you continue to see OOM in step 1, do what I've done for `group deployment tests` in the above PR. Essentially split them into their own `testlist-arm-yourtests.txt`.

darshanhs90:
@balajikris Ran the Build job like 4-5 times after pulling the latest code and it passed with green all the time

balajikris:
@darshanhs90 -- great, that's good to know. I'm also constantly monitoring the health of all builds here: https://travis-ci.org/Azure/azure-xplat-cli/builds and will continue to do so for a while. I think we're good for now. Finally - thank you for the patience around this issue, for diligently following up on the tests and trying to get them to pass 🏅

PR #3598, opened 2017-05-26
Removing "azure mobile" commands

pvbchak:
We have deprecated our service and this PR will remove corresponding CLI commands.

* Removed "Azure Mobile" commands


balajikris:
1. Did you also remove command help text from "plugins.arm.json" (or plugins.asm.json). 
    * Can you perform simple sanity checks like typing out this command no longer displays any messages related to this command
2. Or would you want to add a message like "this command has been removed since the service was deprecated" if the user tries to use your commands ?

balajikris:
@pvbchak - can you please respond to the above question. I'll merge depending on how you think is best to let customers know this feature is now gone.

pvbchak:
@balajikris, I have removed command help text in plugins.asm.json in latest commit. We have announced Mobile Service transition about 1.5 years ago and transitioned all customers to App Services now. So it is safe to remove "azure mobile" commands instead of "marking command has been removed since the service was deprecated".

PR #3599, opened 2017-05-26
Update Compute Package Version

huangpf:
* Update Compute SDK Package to ```2.0.0-preview``` with new API version


huangpf:
All tests passed locally, but failed here. Why?

```
...
  . arm appserviceplan delete should work: 274ms

  855 passing (11m)
  32 pending

End:3:31:22 PM
```

huangpf:
@balajikris

huangpf:
retest this please

balajikris:
@huangpf - I think I've fixed the out-of-memory exception in tests with this PR: https://github.com/Azure/azure-xplat-cli/pull/3600 . Please get the latest changes from `dev` and retest.

balajikris:
@huangpf - I wonder why you have to skip any iothub tests, as they're passing in `dev` branch. Could you explain?

huangpf:
@balajikris It doesn't pass in my local run. I re-enable it for CI test now. Let's see if it passes.

huangpf:
Setting this will make the test pass locally:
```
set AZURE_SUBSCRIPTION_ID=
```

balajikris:
@huangpf - the trick is, if it isn't your test or if it isn't something that shouldn't be affected by your changes, just send a PR and see if the CI passes. If it does, lets not worry about local failures as your environment might not be set up for the entire suite. I suppose thats the best we can do when multiple teams work on the same project. Anyway, all good now 👍

PR #3600, opened 2017-05-30
Travis CI improvements

balajikris:
1. Splitting test commands invoked from travis CI into arm and asm tests. While the breakdown makes sense logically, the intention behind is to shutdown and restart node process inbetween to see if this reduces the memory pressure enough to resolve the intermittent out-of-memory problems we've been observing.
1. upgrade CI environment from `Precise` to `Trusty`. This should get us VMs with 7.5G memory (See [travis-documentation](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments) and [announcement](http://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/))
1. pass --max_old_space_size to increase V8's default memory limit constraints. (See [this](https://github.com/nodejs/node/issues/13018), [this](https://github.com/nodejs/node/issues/10137) and [this](https://stackoverflow.com/questions/38558989/node-js-heap-out-of-memory))
1. split out group deployment tests, which seems to be the exact place where every OOM seems to happen.
1. Better diagnostics: display available system memory before running tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+dist: trusty
+sudo: required
+
 language: node_js
 
 node_js:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ install:
 script:
   - npm -s run-script jshint
   - npm -s run-script unit-arm
+  - npm -s run-script unit-arm-deployment
   - npm -s run-script unit
 
 after_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,9 @@ install:
   - npm install
 
 script:
-  - npm test
+  - npm -s run-script jshint
+  - npm -s run-script unit-arm
+  - npm -s run-script unit
 
 after_script:
   - echo "========== test log: ============"
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@ install:
   - npm --version
   - npm install
 
+before_script:
+  - echo "======== startup memory info ========"
+  - free -h -t
+
 script:
   - npm -s run-script jshint
   - npm -s run-script unit-arm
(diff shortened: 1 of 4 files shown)


balajikris:
The failure is consistently reproducible in `Trusty` whereas it was intermittent in `Precise`. Notice that all commits after moving upto `Trusty` have hit the failure.

balajikris:
@yugangw-msft - could you please see if these changes are okay.

I think I've made a bunch of improvements to our CI as listed above. If you observe commit by commit, you could see OOM popping up on the 2nd and 3rd commits. It became reliably reproducible after I upgraded Travis to use `Trusty`, keep in mind it was still present in the older `Precise` systems but was highly intermittent.
By the 4th and 5th commits, you'd notice we are stable again.

I'm now throwing in extra diagnostics so that, if we run into this in future, we can diagnose the issue more effectively.
~~~
$ free -h -t
                total       used       free     shared    buffers     cached
Mem:          7.3G       2.0G       5.3G       191M        58M       1.3G
-/+ buffers/cache:       674M       6.6G
Swap:           0B         0B         0B
Total:        7.3G       2.0G       5.3G
~~~

.travis.yml
```diff
@@ -11,8 +14,15 @@ install:
   - npm --version
   - npm install
 
+before_script:
```
yugangw-msft: Suggest comment out this section, as it is for diagnostics only

.travis.yml
```diff
@@ -11,8 +14,15 @@ install:
...
+  - echo "======== startup memory info ========"
+  - free -h -t
+
 script:
-  - npm test
+  - npm -s run-script jshint
+  - npm -s run-script unit-arm
+  - npm -s run-script unit-arm-deployment
```
yugangw-msft: add a comment, that we do it to avoid OOM error, so that people will not merge it back in future

balajikris: Done -- thanks!

PR #3601, opened 2017-05-31
Replaced ExpressRoutes' Circuit Peerings with generated version

number213:
Replaced ExpressRoutes' Circuit Peerings with generated version & re-recorded tests

* Rewrote implementation for ExpressRoute Circuit Peerings


balajikris:
@number213 - is it fully autogen'd? Since you guys are now generating stuff, could you please also list what parts are generated and what parts are hand-written (if any), in your PRs? That'll help us review.

number213:
@balajikris will do.

- `ExpressRoutes-expressRouteCircuitPeerings._js` is fully generated
- `expressRoute._js` is handwritten helper. I would remove it but `getCircuit` is used by another cmdlet

balajikris:
Got it - That explains what's going on here. Thanks

PR #3607, opened 2017-06-06
Add ASM Compute VM Commands: Delete/Redeploy/Start/Restart/GetRDP/Show

huangpf:
* 
* 
*


balajikris:
title sounds too generic, its not clear what commands were added. please add a description w/ list of commands added so that the release notes will accurately reflect the changes

Issue #3605, opened 2017-06-01
npm install -g azure-cli fails with npm@5


chrmoritz:
CLI Version: **0.10.13**
OS Type: Mac
Installation via:  both **npm** and **brew** are affected

Mode: (installation fails)

Environment: (installation fails)

Description:
`npm install -g azure-cli` fails when using npm@5 (installed either with `nvm use 8` or the wip  https://github.com/Homebrew/homebrew-core/pull/14085) if not run as root with:
```
npm ERR! path /Users/chris/Desktop/test/azure-xplat-cli-0.10.12-May2017/node_modules/.staging/easy-table-5fa4728c/lib/table.js
npm ERR! code EACCES
npm ERR! errno -13
npm ERR! syscall open
npm ERR! Error: EACCES: permission denied, open '/Users/chris/Desktop/test/azure-xplat-cli-0.10.12-May2017/node_modules/.staging/easy-table-5fa4728c/lib/table.js'
npm ERR!  { Error: EACCES: permission denied, open '/Users/chris/Desktop/test/azure-xplat-cli-0.10.12-May2017/node_modules/.staging/easy-table-5fa4728c/lib/table.js'
npm ERR!   cause: 
npm ERR!    { Error: EACCES: permission denied, open '/Users/chris/Desktop/test/azure-xplat-cli-0.10.12-May2017/node_modules/.staging/easy-table-5fa4728c/lib/table.js'
npm ERR!      errno: -13,
npm ERR!      code: 'EACCES',
npm ERR!      syscall: 'open',
npm ERR!      path: '/Users/chris/Desktop/test/azure-xplat-cli-0.10.12-May2017/node_modules/.staging/easy-table-5fa4728c/lib/table.js' },
npm ERR!   errno: -13,
npm ERR!   code: 'EACCES',
npm ERR!   syscall: 'open',
npm ERR!   path: '/Users/chris/Desktop/test/azure-xplat-cli-0.10.12-May2017/node_modules/.staging/easy-table-5fa4728c/lib/table.js',
npm ERR!   parent: 'azure-cli' }
npm ERR! 
npm ERR! Please try running this command again as root/Administrator.
```

Steps to reproduce:
1) `nvm use 8` or build https://github.com/Homebrew/homebrew-core/pull/14085 from source
2) verify that `npm -v` prints v5.x.x
3) `npm install -g azure-cli` or just `npm install easy-table@0.0.1`

Further informations:

[...]

PR #3608, opened 2017-06-06
upgrade easy table dependency

balajikris:
* Fixed bug https://github.com/Azure/azure-xplat-cli/issues/3605. The issue is that, xplat-cli fails to install with npm5. The root cause was we're using an old version of `easy-table` package that has a bug with file system permissions and was later fixed. This PR updates the version of `easy-table` to a more recent one.

### DIFF
--- a/lib/util/logging.js
+++ b/lib/util/logging.js
@@ -130,25 +130,17 @@ log.table = function (level, data, transform) {
     log.log(level, 'table', data);
   } else {
     var table = new Table();
-    table.LeftPadder = Table.LeftPadder;
+    table.leftPadder = Table.leftPadder;
     table.padLeft = Table.padLeft;
-    table.RightPadder = Table.RightPadder;
+    table.rightPadder = Table.rightPadder;
     table.padRight = Table.padRight;
 
     if (data && data.forEach) {
-      data.forEach(function (item) { transform(table, item); table.newLine(); });
+      data.forEach(function (item) { transform(table, item); table.newRow(); });
     } else if (data) {
       for (var item in data) {
         transform(table, item);
-        table.newLine();
-      }
-    }
-
-    for (var i in table.lines) {
-      for (var column in table.columns) {
-        if (!table.lines[i].hasOwnProperty(column)) {
-          table.lines[i][column] = '';
-        }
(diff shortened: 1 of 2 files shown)


balajikris:
@yugangw-msft - can you please review this. There aren't any tests and `easy-table` had changed a lot since their initial version which we were using. 

I tested a few commands manually `azure account list` and `azure site list` and it prints the table as expected. Any other commands I should try out? FWIW, I'll also test on mac and use npm5 to install dependencies, just to make sure.

lib/util/logging.js
```diff
@@ -130,25 +130,17 @@ log.table = function (level, data, transform) {
...
-      data.forEach(function (item) { transform(table, item); table.newLine(); });
+      data.forEach(function (item) { transform(table, item); table.newRow(); });
     } else if (data) {
       for (var item in data) {
         transform(table, item);
-        table.newLine();
-      }
-    }
-
-    for (var i in table.lines) {
```
balajikris: there is no longer a `table.lines` api. there is a `table.rows` but that doesn't have the same structure as expected below anymore. That makes this dead code (since it wont execute as there is no table.lines) and I could not find any perceivable difference without this, so, I'm removing this completely

lib/util/logging.js
```diff
@@ -130,25 +130,17 @@ log.table = function (level, data, transform) {
     log.log(level, 'table', data);
   } else {
     var table = new Table();
-    table.LeftPadder = Table.LeftPadder;
+    table.leftPadder = Table.leftPadder;
```
balajikris: api was changed

lib/util/logging.js
```diff
@@ -130,25 +130,17 @@ log.table = function (level, data, transform) {
...
     table.padLeft = Table.padLeft;
-    table.RightPadder = Table.RightPadder;
+    table.rightPadder = Table.rightPadder;
     table.padRight = Table.padRight;
 
     if (data && data.forEach) {
-      data.forEach(function (item) { transform(table, item); table.newLine(); });
+      data.forEach(function (item) { transform(table, item); table.newRow(); });
```
balajikris: another api change

PR #3614, opened 2017-06-09
cloud console: fill in isMRRT field when being invoked with raw tokens

yugangw-msft:
All tokens to access azure resource can be acquired using multi-resource-refresh-tokens, hence this filed should `true`


yugangw-msft:
Fix https://github.com/Azure/azure-cli/issues/3641

PR #3617, opened 2017-06-15
Replaced tests for Application Gateways with autogenerated ones

number213:
All files except `arm.network.application-gateway-custom-tests.js` are autogenerated. "Custom tests" file is for tests that are too complex for generator.


balajikris:
thanks for the clear description 👍

Issue #3624, opened 2017-06-22
While Creating an Azure Active Directory (AAD) Application using Azure CLI getting error


anandhs191:
CLI Version: 0.10.14
OS Type:  Win 10
Installation via:  npm / brew / win-or-mac installer / docker / github repo

Mode: **ARM"

Environment: **AzureCloud**

Description:
azure ad app create --name "Service Principal for BOSH" --password "PASSWORD" --home-page "http://BOSHAzureCPI" --identifier-uris "http://BOSHAzureCPI" command has error

info:    Executing command ad app create
error:   Cannot find module 'is-stream'
error:   Error information has been recorded to C:\Users\USERNAME\.azure\azure.err
error:   ad app create command failed

Error stack trace:

{ Error: Cannot find module 'is-stream'
  <<< async stack >>>
  at exports_createApplication__1 (C:\Program Files (x86)\Microsoft SDKs\Azure\CLI\lib\commands\arm\ad\adUtils.js:58:17)  at __1 (C:\Program Files (x86)\Microsoft SDKs\Azure\CLI\lib\commands\arm\ad\ad.application.js:47:33)
  <<< raw stack >>>
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (C:\Program Files (x86)\Microsoft SDKs\Azure\CLI\node_modules\azure-graph\node_modules\ms-rest\lib\serialization.js:7:18)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
  stack: [Getter/Setter],
  code: 'MODULE_NOT_FOUND',
  __frame:
   { name: 'exports_createApplication__1',
     line: 28,
     file: 'C:\\Program Files (x86)\\Microsoft SDKs\\Azure\\CLI\\lib\\commands\\arm\\ad\\adUtils.js',
     prev:
      { name: '__1',
        line: 46,
        file: 'C:\\Program Files (x86)\\Microsoft SDKs\\Azure\\CLI\\lib\\commands\\arm\\ad\\ad.application.js',
        prev: undefined,
        calls: 1,
active: false,
        offset: 1,
        col: 32 },
     calls: 0,
     active: true,
[...]

PR #3629, opened 2017-06-28
[Build Infrastructure] Update windows build script to use..

balajikris:
.. a recent stable version of npm. The script downloads npm 4.6.0 and
unzips it to a temp location and uses it. This ensures our node_modules
folder is deduped. Ensuring we're on a recent stable version of npm
prevents bugs like #3624 which arise because dependency management
software (npm in this case) wasn't updated in a while.

### DIFF
--- a/tools/windows/scripts/prepareRepoClone.cmd
+++ b/tools/windows/scripts/prepareRepoClone.cmd
@@ -10,7 +10,7 @@
 chcp 850 
 
 set NODE_VERSION=6.10.0
-set NPM_VERSION=1.3.17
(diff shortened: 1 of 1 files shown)


balajikris:
@yugangw-msft - this is a proper fix for the issue we encountered during release of 0.10.14. At that time we worked around by installing npm@4.6.0 on the build machine but then the infra team reverted that as it broke powershell team. We now don't rely on npm from the build machine but rather ensure we have an updated npm copy ourselves. This script does just that.

PR #3638, opened 2017-07-12
Replaced tests for Traffic Manager with autogenerated ones

number213:
* Regenerated Traffic Manager using the latest swagger (no new features, just updated descriptions)
* Updated CLI interaction (removed duplicate import)
* Updated generatorUtils (fix for tests, added new `list` method)
* Replaced tests for Traffic Manager with autogenerated ones

All files are autogenerated except:
* lib/util/generatorUtils.js
* lib/util/interaction.js


number213:
Hi, @balajikris, I fixed merge conflicts

lib/util/interaction.js
```diff
@@ -317,7 +316,7 @@ __.extend(Interactor.prototype, {
     if(this.cli.output.format().json) {
       this.cli.output.json(outputData);
     } else {
-      humanOutputGenerator(outputData);
+      humanOutputGenerator(outputData, this.cli);
```
balajikris: I don't think this is right. `formatOutput` is used in thousands of places in this product and `humanOutputGenerator` is passed in as a function that expects 1 argument, not 2. Keep in mind this is a common lib\utils which is used across the product. touching this for your singular purpose is not good.

Also,  I'm not sure why you should change product code when the PR is about replacing "Tests for networking".

lib/util/generatorUtils.js
```diff
@@ -2,44 +2,89 @@ var utils = require('./utils');
 var util = require('util');
 var $ = utils.getLocaleString;
 
-exports.traverse = function(obj, indent) {
-  if(!indent) {
+function getHumanReadableFromCamelCase (inName) {
```
balajikris: 1. why do you need to change product code for updating tests.
2. is this module used only by the networking team?

lib/util/interaction.js
number213: We introduced additional tests that run without `--json` parameter, but for some reason we couldn't pass cli object to `humanOutputGenerator` (it works OK in CLI itself), so I added that tweak. We found another workaround that applies only to our code, so everything should be fine now.

lib/util/generatorUtils.js
number213: 1. See comment in another thread + we added new method for list output that is used in code generator
2. Yes, `generatorUtils` is used only by our team

number213:
@balajikris, addressed your feedback

lib/util/interaction.js
```diff
@@ -18,7 +18,6 @@ var __ = require('underscore');
 var tty = require('tty');
 var fs = require('fs');
 var util = require('util');
-var tty = require('tty');
```
balajikris: why? this will break hugely at runtime as this is being used [here](https://github.com/Azure/azure-xplat-cli/pull/3638/files#diff-81362eaa9090190f40d46da5b1d5735fR35)

lib/util/generatorUtils.js
```diff
@@ -1,55 +1,140 @@
...
+  if (!inName) {
+    return inName;
+  }
+
+  var varName = inName;
```
balajikris: `let` or `const` (if not written to again) instead of `var`

lib/util/generatorUtils.js
```diff
@@ -1,55 +1,140 @@
...
+        }
+        return true;
+      });
+
+      if (matched) {
+        outName += matchedAbbr;
+        i = i + matchedAbbr.length;
+      } else {
+        var j = i + 1;
+        while ((j < varName.length) && varName[j] == varName[j].toLowerCase()) {
```
balajikris: `===`

lib/util/interaction.js
number213: It's a duplicate (see three line 18)

balajikris: Nice !

lib/util/generatorUtils.js
number213: Is ES6 allowed? I remember I saw this commit couple of months ago which stated that jshint prevents from using it.
https://github.com/Azure/azure-xplat-cli/commit/93350a4b73d6dd1d7a8ca95b5252231e2e11f97f

balajikris: Hmm.. Min version of node required for this app is `6.10.1` which supports ES6. but it seems like some infra work remains to be done to make jshint accept es6? given that, let this go as such for now.

lib/util/generatorUtils.js
number213: Done

balajikris:
@number213 - Thanks for the changes. LGTM, will merge once CI passes.

PR #3647, opened 2017-08-01
Replaced the rest of Network tests with autogenerated ones

number213:
This PR concludes replacement of CLI Tests. 

* Network test list is moved to a separate file
* Everything else is autogenerated


balajikris:
Having small manageable PRs would be nice. Regardless of whether the change is generated by a machine or by a human, the least amount of information I'd like to know is "what files are touched". 
Say, for e.g. - I'd like to know whether a common or critical module such as utils.js or azure.js was touched. In a PR like this, github does not even load the list of files changed. Viewing the content diffs is simply out of the question. Since I could neither understand the list of files changed, nor the changes themselves, I'm not comfortable signing off on this. 

Please split it into multiple PRs. Thanks.

PR #3648, opened 2017-08-02
updating README for npm5 commands

carlowahlstedt:
Based on the following error, updated to use the recommended verify command. I didn't realize until I was through several of the commands that I wanted to the next step. As such, I would also recommend moving the install from npm step to be the first install step since that's what most would likely take.

$ npm cache clear -g
npm ERR! As of npm@5, the npm cache self-heals from corruption issues and data extracted from the cache is guaranteed to be valid. If you want to make sure everything is consistent, use 'npm cache verify' instead.
npm ERR! 
npm ERR! If you're sure you want to delete the entire cache, rerun this command with --force.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/carlo.wahlstedt/.npm/_logs/2017-08-02T13_23_36_247Z-debug.log

* 
* 
*

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,9 @@ Please take a look at the list of supported environment variables over [here](ht
 - Uninstall the previously installed CLI
    - If you installed via MSI, then uninstall the windows MSI. For mac installer `sudo azure-uninstall -g`
    - If you installed via npm then execute: `npm uninstall -g azure-cli`
-- Clear the global cache: `npm cache verify`
+- Clear the global cache: 
+   - If npm4.0 or before: `npm cache clear -g`
+   - If npm5.0 or greater `npm cache verify`
 - Delete the .streamline folder from the User’s home profile `C:\Users\<username>\.streamline` | `~/.streamline`, if present.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,7 +44,7 @@ Please take a look at the list of supported environment variables over [here](ht
 - Uninstall the previously installed CLI
    - If you installed via MSI, then uninstall the windows MSI. For mac installer `sudo azure-uninstall -g`
    - If you installed via npm then execute: `npm uninstall -g azure-cli`
-- Clear the global cache: `npm cache clear –g`
+- Clear the global cache: `npm cache verify`
```
balajikris: This will only work if the user is on npm5.0 or greater. Whereas the previous command tells you how to resolve the problem, using this new command against a older version of npm would not be helpful. I'd prefer having both the commands, with a comment about which command to use depending on the version of npm installed on the machine. Does that sound agreeable?

carlowahlstedt:
@balajikris excellent point. I've made changes reflecting this.

balajikris:
Thanks for the contribution, @carlowahlstedt 👍

PR #3652, opened 2017-08-03
[Network] Replaced tests for Express Routes with autogenerated ones

number213:
Everything is autogenerated except:
* `test/commands/arm/network/arm.network.express-route-circuit-tests.js` (contains test scenarios that are too complex for autogenerator)


number213:
Resolved conflicts

PR #3653, opened 2017-08-07
[Network] Improved list output & fixed nowait in Application Gateway

number213:
Re-generated Network cmdlets which resulted in:

* Updated `list` method to include more properties in output
* Fixed duplicate shorthands in VPN Gateways
* Fixed nowait support in ApplicationGateway

* Network
  * Improved output of list commands
  * Application Gateway
    * Fixed `--nowait` support 
  * Virtual Network Gateway
    * The shorthand for `--enable-active-active-feature` was changed to `-r` as it was conflicting with the shorthand for `--resource-group`


balajikris:
Hey @number213 -
your description says "regenerated". So, can i assume all changes were auto-generated and not manually touched?

number213:
@balajikris yes, sorry, forgot to write that. Everything is autogenerated except `lib/util/generatorUtils.js`

balajikris:
Thanks @number213

PR #3655, opened 2017-08-15
[Network] Added CheckConnectivity method to Network Watcher

number213:
All files are generated except `test/util/preinstalledEnv.js`

* Added CheckConnectivity method to Network Watcher
* Moved autogenerated tests into separate folder to mirror the same approach with command files

* Network
  * Added CheckConnectivity method


balajikris:
LGTM

PR #3668, opened 2017-10-31
[Do not merge yet] Added support for DNS CAA records

EvgenyAgafonchikov:
~Depends on: https://github.com/Azure/azure-sdk-for-node/pull/2306~

Added DNS CAA tests; Updated recordings

* Added support for DNS CAA records


lib/commands/arm/network/zoneFile._js
```diff
@@ -641,6 +659,10 @@ __.extend(ZoneFile.prototype, {
         rr += util.format('%s %s %s %s', record.priority, record.weight, record.port, record.target);
         if (!utils.stringEndsWith(record.target, '.', true)) rr += '.';
         break;
+      case constants.dnsZone.CAA:
+        rr += util.format('%s %s %s', record.flags, record.tag, record.value);
+        if (!utils.stringEndsWith(record.value, '.', true)) rr += '.';
```
nkindberg: We should not be appending any "." to CAA values - customer input should be taken as-is

lib/commands/arm/network/network._js
```diff
@@ -461,6 +461,10 @@ exports.init = function (cli) {
     .option('-w, --weight [weight]', $('the weight attribute'))
     .option('-o, --port [port]', $('the port'))
     .option('-u, --target [target]', $('the target attribute\n' +
+      '\nRecord type CAA'))
+    .option('-i, --caa-flags [flags]', $('the flags attribute'))
```
nkindberg: Our PM spec'd this as below - no dash in the middle. Please use naming without the dash to ensure we are consistent across both CLI 1.0 and 2.0

--caaflags
--caatag
--caavalue

lib/plugins.arm.json
```diff
@@ -113698,7 +113698,34 @@
                       "bool": true,
                       "short": "-u",
                       "long": "--target",
-                      "description": "the target attribute\n\nRecord type TXT"
+                      "description": "the target attribute\n\nRecord type CAA"
+                    },
+                    {
+                      "flags": "-i, --flags [flags]",
```
nkindberg: Is this for internal use, or this is exposed in CLI to customer?

If exposed to customer, let's make sure to be consistent with PM spec

--caaflags
--caatag
--caavalue

lib/plugins.arm.json
```diff
@@ -113894,7 +113921,34 @@
...
+                      "description": "the tag attribute"
+                    },
+                    {
+                      "flags": "-m, --value [value]",
+                      "required": 0,
+                      "optional": -13,
+                      "bool": true,
+                      "short": "-m",
+                      "long": "--value",
+                      "description": "the value attribute\n\nRecord type TXT"
```
nkindberg: Mentions record type TXT here - this is for CAA?

lib/plugins.arm.json
EvgenyAgafonchikov: this file is generated, so it would contain values from the real parameter names

lib/plugins.arm.json
EvgenyAgafonchikov: this is added to the last CAA parameter to achieve adding of categories headers. i.e. this would look like this in CLI help message:
```
Record type CAA
    <CAA parameters>
Record type TXT
    <TXT parameters>
```

number213:
Sinon requires package called `type-detect` which has been recently updated to version 4.0.4 which fails during import. See this build (https://travis-ci.org/Azure/azure-xplat-cli/builds/299632371)

Specifying version as 4.0.3 in `package.json` seems to solve the problem.

number213:
Hi, @amarzavery, could you please look at this PR? Evgeny is OOF, so I can't change the title to remove "do not merge" tags.
Also, regarding previous comment on package dependencies, it has been fixed (https://github.com/chaijs/type-detect/pull/120) so I reverted my changes to package.json.

PR #3679, opened 2018-02-07
support msi based account

yugangw-msft:
Align to the plan that cloud shell goes to msi model for creds management


lib/util/authentication/msiAuth.js
```diff
@@ -0,0 +1,57 @@
...
+
+function MsiTokenCredentials(resource, port) {
+  this._resource = resource;
+  this._port = port;
+}
+
+util.inherits(MsiTokenCredentials, TokenCredentials);
+
+//'retrieveTokenFromCache' is confusing, but for the sake of conforming to the interface.
+MsiTokenCredentials.prototype.retrieveTokenFromCache = function (callback) {
```
amarzavery: Is it possible to add this class in ms-rest-azure. We have a bunch of MSIVmTokenCredentials and MSIAppServiceTokenCredentials.
You can take a look at this class over here https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest-azure/lib/credentials/msiVmTokenCredentials.js#L17

I would recommend changing the minimum dependency of ms-rest-azure to `^2.5.2` over here https://github.com/Azure/azure-xplat-cli/blob/dev/package.json#L93.

This will ensure that you get the latest version of ms-rest-azure until today all the time.

yugangw-msft: Thanks. Shame on me for forgetting this good support from sdk

